PR #177, opened 2019-04-25
Find path to packages by looking for package.xml (fixes #175, fixes #181)

### LINKED ISSUE #175: Package database construction fails for unusual ROS_PACKAGE_PATH env vars
Usually, each element in the `ROS_PACKAGE_PATH` corresponds to a single package. ROSWire relies on this observation to find the location of each package and to parse its contents. However, in some cases, elements within the `ROS_PACKAGE_PATH` may correspond to a directory that contains several packages. ROSWire should accommodate for this case.

The best way to handle this issue is to do the following for each element in `ROS_PACKAGE_PATH`:
* check if the directory exists.
* check if the directory contains a `package.xml` file. If so, it's a package directory. If not, the element may correspond to a directory that hosts multiple ROS packages. In that case, recursively find all ROS packages within each of its subdirectories.

### LINKED ISSUE #181: Add find method to FileProxy
Should accept the following arguments (as a minimum):
* the absolute path of the directory that should be searched.
* the name of the file.

afsafzal:


src/roswire/definitions/package.py
```diff
@@ -90,7 +90,22 @@ def paths(shell: ShellProxy) -> List[str]:
...
+        package_paths: List[str] = path_str.strip().split(':')
+        paths: List[str] = []
+        for path in package_paths:
+            print("P: {}".format(path))
+            code, _, _ = shell.execute('test -f {}/package.xml'.format(path))
+            print("code: {}".format(code))
+            if code != 0:
+                paths.append(path)
+            else:
+                code, dirs_str, _ = shell.execute('ls -d {}/*'.format(path))
```
ChrisTimperley: Use `shlex.quote`.

src/roswire/definitions/package.py
```diff
@@ -90,7 +90,22 @@ def paths(shell: ShellProxy) -> List[str]:
         code, path_str, duration = shell.execute('echo "${ROS_PACKAGE_PATH}"')
         if code != 0:
             raise Exception("unexpected error when fetching ROS_PACKAGE_PATH")
-        paths: List[str] = path_str.strip().split(':')
+        package_paths: List[str] = path_str.strip().split(':')
+        paths: List[str] = []
+        for path in package_paths:
+            print("P: {}".format(path))
+            code, _, _ = shell.execute('test -f {}/package.xml'.format(path))
```
ChrisTimperley: `shlex.quote`.

src/roswire/definitions/package.py
```diff
@@ -90,7 +90,22 @@ def paths(shell: ShellProxy) -> List[str]:
...
+            print("code: {}".format(code))
+            if code != 0:
+                paths.append(path)
+            else:
+                code, dirs_str, _ = shell.execute('ls -d {}/*'.format(path))
+                print("dirs_str: {}".format(dirs_str))
+                if code != 0:
+                    raise Exception("unexpected error when listing directories")
+                dirs: List[str] = path_str.strip().split('\n')
+                package_paths.extend(dirs)
```
ChrisTimperley: It shouldn't be possible to modify a list whilst iterating over it. Maybe switch to using a Queue instead?

src/roswire/definitions/package.py
ChrisTimperley: Or, better yet, use the FileProxy API to do this.

src/roswire/definitions/package.py
ChrisTimperley: Use the FileProxy API.

src/roswire/definitions/package.py
```diff
@@ -90,7 +91,25 @@ def paths(shell: ShellProxy) -> List[str]:
         code, path_str, duration = shell.execute('echo "${ROS_PACKAGE_PATH}"')
         if code != 0:
             raise Exception("unexpected error when fetching ROS_PACKAGE_PATH")
-        paths: List[str] = path_str.strip().split(':')
+        package_paths: List[str] = path_str.strip().split(':')
+        paths: List[str] = []
+        for path in package_paths:
+            cmd = f'test -f {shlex.quote(path)}/package.xml'
```
ChrisTimperley: This is better, but it makes more sense to add `FileProxy` to the args for this function and to use its interface to perform this operation.

src/roswire/definitions/package.py
```diff
@@ -90,7 +91,25 @@ def paths(shell: ShellProxy) -> List[str]:
...
+            code, _, _ = shell.execute(cmd)
+            if code == 0:
+                paths.append(path)
+            else:
+                cmd = f'ls -d {shlex.quote(path)}/*'
```
ChrisTimperley: Same as above.

src/roswire/definitions/package.py
```diff
@@ -90,7 +91,25 @@ def paths(shell: ShellProxy) -> List[str]:
...
+                code, dirs_str, _ = shell.execute(cmd)
+                if code != 0:
+                    # Not a directory
+                    continue
+                dirs: List[str] = dirs_str.strip().split('\n')
+                for d in dirs:
+                    cmd = f'test -f {shlex.quote(d)}/package.xml'
```
ChrisTimperley: Same as above.

src/roswire/definitions/package.py
```diff
@@ -83,7 +83,7 @@ def to_dict(self) -> Dict[str, Any]:
 
 class PackageDatabase(Mapping[str, Package]):
     @staticmethod
-    def paths(shell: ShellProxy) -> List[str]:
+    def paths(shell: ShellProxy, file_proxy: FileProxy) -> List[str]:
```
ChrisTimperley: Just `files` is okay.

src/roswire/definitions/package.py
```diff
@@ -82,15 +83,28 @@ def to_dict(self) -> Dict[str, Any]:
...
+            if file_proxy.isfile(f'{path}/package.xml'):
+                paths.append(path)
+            else:
+                try:
+                    dirs: List[str] = file_proxy.listdir(path, absolute=True)
+                except OSError:
+                    # Not a directory
+                    continue
+                for d in dirs:
+                    if file_proxy.isfile(f'{d}/package.xml'):
```
ChrisTimperley: This doesn't work for cases where `package.xml` is nested more than one subdirectory deep. Using a queue is probably better.

src/roswire/definitions/package.py
```diff
@@ -82,15 +83,28 @@ def to_dict(self) -> Dict[str, Any]:
...
             raise Exception("unexpected error when fetching ROS_PACKAGE_PATH")
-        paths: List[str] = path_str.strip().split(':')
+        package_paths: List[str] = path_str.strip().split(':')
+        paths: List[str] = []
+        for path in package_paths:
+            if file_proxy.isfile(f'{path}/package.xml'):
+                paths.append(path)
+            else:
+                try:
+                    dirs: List[str] = file_proxy.listdir(path, absolute=True)
```
ChrisTimperley: You could just extend a queue with all of the subdirectories at this point.

src/roswire/definitions/package.py
afsafzal: I tried that. It takes a very long time to traverse all subdirectories. Another way is to just use `find` to list all `package.xml` files. Do you think that's a better solution?

src/roswire/definitions/package.py
afsafzal: The comment above.

src/roswire/definitions/package.py
ChrisTimperley: That sounds good. It would be great if we could add an interface for `find` to `FileProxy`.

ChrisTimperley: I've created a new issue: https://github.com/ChrisTimperley/roswire/issues/181

src/roswire/proxy/file.py
```diff
@@ -186,6 +186,28 @@ def islink(self, path: str) -> bool:
...
 
+    def find(self,
+             path: str,
+             filename: str
+             ) -> List[str]:
+        """
+        Returns a list of files that match the provided filename
+        in a given directory recursively.
+
+        Parameters:
```
ChrisTimperley: This is good, but not quite the right format: https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html

src/roswire/definitions/package.py
```diff
@@ -94,17 +94,13 @@ def paths(shell: ShellProxy, file_proxy: FileProxy) -> List[str]:
...
-                    dirs: List[str] = file_proxy.listdir(path, absolute=True)
-                except OSError:
-                    # Not a directory
-                    continue
-                for d in dirs:
-                    if file_proxy.isfile(f'{d}/package.xml'):
-                        paths.append(d)
+            try:
+                all_packages = files.find(path, 'package.xml')
+            except OSError:
```
ChrisTimperley: https://docs.python.org/3/library/contextlib.html#contextlib.suppress

src/roswire/proxy/file.py
```diff
@@ -186,6 +186,28 @@ def islink(self, path: str) -> bool:
...
+            filename: the name of the file to match.
+
+        Raises:
+            OSError: if find operation is not successful.
+        """
+        cmd = f'find {shlex.quote(path)} -name {shlex.quote(filename)}'
+        code, output, duration = self.__shell.execute(cmd)
+        if code != 0:
+            raise OSError(f"Find operation was unsuccessful on {path}")
+        paths: List[str] = output.replace('\r', '').split('\n')
```
ChrisTimperley: Will this return the absolute file path? It might be worth writing a test for this method.

afsafzal: I think it returns `path`+subdirectories. So if `path` is absolute, the returned paths are absolute too.

src/roswire/definitions/package.py
afsafzal: But I don't want to suppress it. I want to continue the for loop if there's an exception.

src/roswire/proxy/file.py
afsafzal: Is this format different than what you've used in other functions of this class?

src/roswire/proxy/file.py
afsafzal: I'll add a test.

src/roswire/definitions/package.py
ChrisTimperley: Ah. Good point.

src/roswire/proxy/file.py
ChrisTimperley: Apparently I've messed up the documentation in this module. Don't worry about it. I'll create an issue to look at the documentation later.

PR #184, opened 2019-04-26
Added example bug to hello world system

afsafzal:
`rosrun rospy_tutorials listener` & `rosbag play example.bag` will result in `listener` node throwing exceptions periodically.


afsafzal:
This is a very stupid bug and it doesn't result in a node crash for the buggy node.

ChrisTimperley:
A bug that crashes the node would be best.

test/hello_world/Dockerfile
```diff
@@ -1,9 +1,31 @@
...
- && apt-get install -y \
+ && apt-get install -y build-essential qtbase5-dev \
       ros-indigo-ros-tutorials \
       ros-indigo-common-tutorials \
       vim \
       wget \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists
+
+RUN mkdir /root/overlay_ws \
```
ChrisTimperley: Can you rename `/root/overlay_fs` to `/ros_ws`? Also, `ROS_WSPACE` could be lifted to before this `RUN` command.

test/hello_world/Dockerfile
```diff
@@ -1,9 +1,31 @@
...
+    && cd /root/overlay_ws \
+    && wstool init src \
+    && cd /root/overlay_ws/src \
+    && wstool set ros_tutorials --git git://github.com/ros/ros_tutorials.git -y \
+    && wstool update 
+
+RUN cd /root/overlay_ws \
```
ChrisTimperley: Also, this could be folded into the `RUN` command above.

ChrisTimperley:
> A bug that crashes the node would be best.

And a C++ example would be even better.

test/hello_world/Dockerfile
```diff
@@ -1,31 +1,34 @@
...
 WORKDIR "${ROS_WSPACE}"
 RUN echo "#!/bin/bash \n\
 set -e \n\
-source \"/opt/ros/indigo/setup.bash\" \n\
+source \"/opt/ros/${ROS_DISTRO}/setup.bash\" \n\
 source \"${ROS_WSPACE}/devel/setup.bash\" \n\
 exec \"\$@\"" > "${ROS_WSPACE}/entrypoint.sh" \
  && chmod +x "${ROS_WSPACE}/entrypoint.sh"
-ENTRYPOINT ["/root/overlay_ws/entrypoint.sh"]
+ENTRYPOINT ["/ros_ws/entrypoint.sh"]
```
ChrisTimperley: Should this be `${ROS_SPACE}`?

afsafzal: If I set it to `ENTRYPOINT ["${ROS_WSPACE}/entrypoint.sh"]` it gives me error that it doesn't recognize `${ROS_WSPACE}`

ChrisTimperley: Interesting. Okay. `/ros_ws/` is fine for now.

test/hello_world/Dockerfile
```diff
@@ -1,9 +1,34 @@
-FROM ros:indigo-ros-base
+FROM ros:indigo
```
ChrisTimperley: Both of these Dockerfiles can be collapsed into one by using a multi-stage build: https://docs.docker.com/develop/develop-images/multistage-build/

```
FROM ros:indigo as base
...

FROM base as buggy
...
```

afsafzal: You mean having one Dockerfile instead of two?

PR #335, opened 2020-03-14
Refactored to use DockerBlade to provide shell and file access

ChrisTimperley:
These changes help to avoid code duplication across several projects (e.g., Darjeeling and BugZoo) and make it easier to optimise the performance of ROSWire, which is currently bottlenecked by its Docker interactions.

For the most part, these changes have no effects on the API. There are two important changes:
1. The `execute` method in `ShellProxy` has been replaced by `run`. There are now also a bunch of other ways to execute commands including `check_call`, `check_output`, and `popen`.
2. More importantly, users must now specify a set of sources when launching or analysing a ROS application. This both (a) avoids relying on the assumption that the Dockerfile has an appropriate entrypoint, which has caused problems with images in the past, and (b) makes it possible to test both `devel` and `install` workspaces (and other exotic configurations).

For example, the following code snippet:
```python
description = rsw.descriptions.load_or_build('roswire/example:mavros')
```

becomes:

```python
sources = ['/opt/ros/indigo/setup.bash', '/ros_ws/devel/setup.bash']
description = rsw.descriptions.load_or_build('roswire/example:mavros', sources)
```

Likewise, the following code snippet:

```python
with rsw.launch('roswire/example:mavros') as system:
```

becomes:

```python
sources = ['/opt/ros/indigo/setup.bash', '/ros_ws/devel/setup.bash']
with rsw.launch('roswire/example:mavros', sources) as system:
```


DebyKatz:
I'm getting an error when trying to import roswire on this branch. ---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
<ipython-input-1-0556d322f6a7> in <module>
----> 1 import roswire

~/.local/share/virtualenvs/overhead-timing-effects-in-ROS-Wl9qFQgJ/lib/python3.6/site-packages/roswire/__init__.py in <module>
      4 
      5 from .version import __version__
----> 6 from .proxy import ShellProxy, ROSProxy
      7 from .system import System, SystemDescription
      8 from .roswire import ROSWire

~/.local/share/virtualenvs/overhead-timing-effects-in-ROS-Wl9qFQgJ/lib/python3.6/site-packages/roswire/proxy/__init__.py in <module>
     26 import time
     27 
---> 28 from .shell import ShellProxy
     29 from .file import FileProxy
     30 from .catkin import CatkinProxy, CatkinToolsProxy, CatkinMakeProxy

ModuleNotFoundError: No module named 'roswire.proxy.shell'

PR #358, opened 2020-05-29
Fix the wrong return in locate function of ROSLaunchManager

afsafzal:


afsafzal:
I'm not sure what's the difference between `filename` and `filename_original`. Should I remove the latter?

ChrisTimperley:
> I'm not sure what's the difference between `filename` and `filename_original`. Should I remove the latter?

Thanks for spotting this! There's a few code smells that led to this issue. The intent behind `filename` and `filename_original` is that `filename` is a mutable variable that should be returned, and `filename_original` is immutable and used for the sake of logging. I messed up the code by a new variable, `path`, into the mix, which removes the need for the `filename_original` and `filename` distinction. I think that a slightly better fix here is to replace all references to `path` with `filename` to restore the original intent.

src/roswire/proxy/roslaunch/roslaunch.py
```diff
@@ -114,12 +114,12 @@ def locate(self, filename: str, *, package: Optional[str] = None) -> str:
                      f' in package [{package}]')
         command = f'rospack find {shlex.quote(package)}'
         try:
-            path = self._shell.check_output(command, stderr=False)
+            filename = self._shell.check_output(command, stderr=False)
```
ChrisTimperley, line 117: Ahh. So this here should be `package_path`.

src/roswire/proxy/roslaunch/roslaunch.py
```diff
@@ -114,12 +114,12 @@ def locate(self, filename: str, *, package: Optional[str] = None) -> str:
...
         except dockerblade.CalledProcessError as err:
             raise exc.PackageNotFound(package) from err
-        path = os.path.join(path, 'launch', filename)
-        if not self._files.isfile(path):
-            raise exc.LaunchFileNotFound(path=path)
+        filename = os.path.join(filename, 'launch', filename_original)
```
ChrisTimperley, line 120: First argument should be `package_path`

PR #359, opened 2020-05-29
Add ROSLaunchController for controlling ROS launch processes (fixes #354)

### LINKED ISSUE #354: Add ROSLaunchController for controlling ROS launch processes
Calling the `launch` method of `ROSLaunchManager` would no longer return `None` and instead return a `ROSLaunchController` instance, allowing the user to interact with the underlying `roslaunch` process. Note that this change is fully  backwards compatible as users can simply choose to ignore the return value as they do now.

At a minimum, the `ROSLaunchController` could expose the following methods:
* `terminate`: terminates the underlying `roslaunch` process
* `close`: alias for `terminate`
* `is_running`: determines whether or not the underlying `roslaunch` process is running.
* `__enter__`: does nothing; used to satisfy requirements of a context-managed object.
* `__exit__`: closes the `roslaunch` via `close`

By adding `__enter__` and `__exit__`, users can use the `with` keyword in conjuction with the `ROSLaunchController` instance to automatically close the launch process at the end of a context:

```python
launch = roscore.roslaunch(...)
with launch:
    # do stuff
# launch is dead!

with roscore.roslaunch(...) as launch:
    # do stuff

# launch is dead!
```

There is also the possibility that the `ROSLaunchController` could expose the logs for its associated `roslaunch` process. This feature would be useful for debugging `roslaunch` processes that do not behave as expected.

ChrisTimperley:


src/roswire/proxy/roslaunch/controller.py
```diff
@@ -0,0 +1,53 @@
...
+                 ) -> None:
+        self.close()
+
+    @property
+    def stream(self) -> Iterator[str]:
+        yield from self.popen.stream()  # type: ignore
+
+    def is_running(self) -> bool:
+        """Checks whether or not this roslaunch process is still running."""
+        return self.popen.finished
```
afsafzal, line 45: I believe this should be `return not self.popen.finished` to make sense.
```suggestion
        return not self.popen.finished
```

src/roswire/proxy/roslaunch/controller.py
```diff
@@ -0,0 +1,53 @@
...
+                 ex_val: Optional[BaseException],
+                 ex_tb: Optional[TracebackType]
+                 ) -> None:
+        self.close()
+
+    @property
+    def stream(self) -> Iterator[str]:
+        yield from self.popen.stream()  # type: ignore
+
+    def is_running(self) -> bool:
```
afsafzal, line 43: I think it makes more sense that here we try to be as close as possible to the `Popen` interface. So instead of `is_running` if we could have `finished` that would be more consistent.

src/roswire/proxy/roslaunch/controller.py
ChrisTimperley: This made me chuckle. Well spotted.

src/roswire/proxy/roslaunch/controller.py
ChrisTimperley: That's fair.

ChrisTimperley: Actually, on second thought, I'm not sure that the `finished` does make as much sense here. What does it mean for a launch group to be finished? The naming is a little ambiguous: Does it mean that the nodes have (abruptly) terminated, or that the process of "launching" those nodes is finished? In any case, we could use turn this into a `@property` named `running` to better match similar interfaces in ROSWire.

PR #360, opened 2020-06-03
Added ability to restrict bag recording to specific topics (fixes #341)

### LINKED ISSUE #341: Add restrict_to_topics to BagRecorder

ChrisTimperley:


src/roswire/proxy/bag.py
```diff
@@ -221,6 +226,8 @@ def start(self) -> None:
                     f'__name:={self.__bag_name}']
             if self.__exclude_topics:
                 args += ['-x', shlex.quote(self.__exclude_topics)]
+            if self.__restrict_to_topics:
+                args += ['-e', shlex.quote(self.__restrict_to_topics)]
```
afsafzal, line 230: I don't think `shlex.quote` here is doing what we want it to do. Right now, let's say for topics `(.*)/move_group/(.*)` it is generating the command as `-e (.*)/move_group/(.*)`, while it should be `-e "(.*)/move_group/(.*)"` to work.

I'm not sure how's best to deal with this. I have the dummy solution of:
```
args += ['-e', '"{}"'.format(shlex.quote(self.__restrict_to_topics))]
```

src/roswire/proxy/bag.py
```diff
@@ -221,6 +226,8 @@ def start(self) -> None:
                     f'__name:={self.__bag_name}']
```
afsafzal, line 226: Having the option `-a` in the arguments overrides option `-e`, and that's why it's not working.

ChrisTimperley: Ah. That makes sense. Thanks for fixing this! :+1:

ChrisTimperley:
The CI is failing, but upon closer look, it seems that `mypy` is finding issues elsewhere in the code that it hasn't identified in the past. Given that, I'll go ahead and merge this PR, and fix those newly discovered issues on another branch.

PR #366, opened 2020-06-10
Added support for name remappings to launch method (fixes #343)

### LINKED ISSUE #343: Add remapping to launch method

ChrisTimperley:


ChrisTimperley:
Parameters with values that are themselves XML documents, such as `robot_description`, can't be written to a legal launch XML. To get around this, we need to use the original `command` that was used to generate the parameter value: https://github.com/ChrisTimperley/roswire/issues/367

PR #370, opened 2020-06-12
Implemented per-node launch prefixes (fixes #347)

### LINKED ISSUE #347: Add launch_prefixes option to launch

ChrisTimperley:


src/roswire/proxy/roslaunch/config/node.py
```diff
@@ -30,6 +30,9 @@ class NodeConfig:
     def full_name(self) -> str:
         return namespace_join(self.namespace, self.name)
 
+    def with_launch_prefix(self, launch_prefix: str) -> 'NodeConfig':
+        return self.with_launch_prefix(launch_prefix)
```
afsafzal, line 34: This is recursive. I assume here we want to create a new `NodeConfig` where the `launch_prefix` is set.

src/roswire/proxy/roslaunch/roslaunch.py
```diff
@@ -184,8 +184,8 @@ def launch(self,
             logger.debug(f'instrumenting launch config: {launch_config}')
 
             if launch_prefixes:
-                m = "individual launch prefixes are not yet implemented"
-                raise NotImplementedError(m)
+                launch_config = \
+                    launch_config.with_launch_prefixes(launch_prefixes)
```
afsafzal, line 188: This is cool, but the use case I'm envisioning will be inefficient with this. Because I want to get the list of the nodes (e.g., all python nodes) automatically, I have to create the `LaunchConfig` beforehand. And then here, we read the launch file again. Maybe it would make sense to allow passing a `LaunchConfig` to this method. What do you think?

ChrisTimperley: I see. That makes sense. We could either add a `launch_from_config` method, or change the `filename` parameter to `filename_or_config`. I think the former makes more sense here.

PR #379, opened 2020-07-08
Added partial implementation of ROS2LaunchManager for launching ROS2 applications (fixes #377)

### LINKED ISSUE #377: Add ROS2LaunchManager for launching ROS2 applications
Should mirror the same interface as ROSLaunchManager: https://github.com/ChrisTimperley/roswire/blob/master/src/roswire/proxy/roslaunch/roslaunch.py

mvkashyap:


ChrisTimperley:
Don't forget to add an entry to the `CHANGELOG` :-)

src/roswire/proxy/roslaunch/ros2launch.py
```diff
@@ -0,0 +1,189 @@
...
+
+from .config import LaunchConfig
+from .controller import ROSLaunchController
+from ... import exceptions as exc
+from .app.app import App
+
+
+@attr.s(eq=False)
+class ROS2LaunchManager:
+    """Provides access to `roslaunch <wiki.ros.org/roslaunch/>`_ for an
```
ChrisTimperley, line 20: Maybe update the docstring to refer to the ROS2 support.

src/roswire/proxy/roslaunch/ros2launch.py
```diff
@@ -0,0 +1,189 @@
...
+            supplied to :code:`roslaunch`.
+
+        Raises
+        ------
+        PackageNotFound
+            If the given package could not be found.
+        LaunchFileNotFound
+            If the given launch file could not be found in the package.
+        """
+    raise NotImplementedError("ROS2 might not be able to read")
```
ChrisTimperley, line 54: No need for message.
```
raise NotImplementedError
```
should be just fine!

src/roswire/proxy/roslaunch/ros2launch.py
```diff
@@ -0,0 +1,189 @@
...
+        -------
+        str
+            The absolute path to the generated XML launch file.
+        """
+    raise NotImplementedError("ROS2 might not be able to write")
+
+    def locate(self,
+               filename: str,
+               *,
+               app: App,
```
ChrisTimperley, line 83: Ah, now I see what you mean :slightly_smiling_face: So, the fix here is for there to be an `_app_instance: AppInstance = attr.ib()` attribute at the top. We can then remove `_shell` and `_files` and replace references to those with `_app_instance.shell` and `_app_instance.files`.

ChrisTimperley:
This looks really good! Thanks for all of your hard work on this. There's a few small changes that we need to make, and then we go ahead and merge.

src/roswire/proxy/roslaunch/ros2launch.py
```diff
@@ -0,0 +1,189 @@
...
+@attr.s(eq=False)
+class ROS2LaunchManager:
+    """Provides access to `roslaunch <wiki.ros.org/roslaunch/>`_ for an
+    associated ROS system. This interface is used to locate, read, and write
+    `launch XML files <http://wiki.ros.org/roslaunch/XML>`_,
+    and to launch ROS nodes using those files.
+    """
+    _shell: dockerblade.shell.Shell = attr.ib(repr=False)
+    _files: dockerblade.files.FileSystem = attr.ib(repr=False)
+
```
ChrisTimperley, line 27: Let's add
```
@classmethod
def for_app_instance(cls, app_instance: 'AppInstance') -> 'ROS2LaunchManager`:
    return ROS2LaunchManager(....)
```

That way we're robust to any changes in the constructor :-) (Btw, if you're curious about these sorts of API choices in general, you should check out Effective Java by Joshua Block. The book is for Java, but many of the same lessons can be applied to other object-oriented languages.)

src/roswire/ros2/ros2.py
```diff
@@ -9,6 +9,7 @@
 from .service_manager import ROS2ServiceManager
 from .state import ROS2StateProbe
 from ..proxy import SystemState
+from ..proxy.ros22launch import ROS2LaunchManager
```
ChrisTimperley, line 12: Looks like a typo? Also, one last change: Can you move that file from `proxy/roslaunch/ros2launch.py` to `ros2/launch.py`?

src/roswire/ros2/ros2.py
```diff
@@ -21,14 +22,17 @@ class ROS2:
     nodes: ROS2NodeManager = attr.ib(init=False)
     services: ROS2ServiceManager = attr.ib(init=False)
     _state_probe: ROS2StateProbe = attr.ib(init=False)
+    launch_manager: ROS2LaunchManager = attr.ib(init=False)
 
     def __attrs_post_init__(self) -> None:
         nodes = ROS2NodeManager.for_app_instance(self.app_instance)
         services = ROS2ServiceManager.for_app_instance(self.app_instance)
         state_probe = ROS2StateProbe.for_app_instance(self.app_instance)
+        launch_manager = ROS2LaunchManager.for_app_instance(self.app_instance)
```
ChrisTimperley, line 31: Just `launch` rather than `launch_manager` is fine.

docs/recipes/inspect_runtime.py
```diff
@@ -37,7 +37,8 @@
         # get topic types
         topic_to_type = ros.topic_to_type
         print(f"Topic to type: {topic_to_type}")
-
+        print("rosnodes:")
```
ChrisTimperley, line 40: We probably want to add a whole bunch of ROS2 recipes at some point, but that's another issue. We can retain the original recipe for now.

src/roswire/ros2/launch.py
```diff
@@ -8,9 +8,9 @@
 from loguru import logger
 import attr
 
-from .config import LaunchConfig
-from .controller import ROSLaunchController
-from ... import exceptions as exc
+from .proxy.roslaunch.config import LaunchConfig
```
ChrisTimperley, line 11: I think this _should_ be `from ..proxy`?

ChrisTimperley:
This looks great. Do you mind running `pytest` to make sure that we're not introducing any regressions? If that passes, I think that we're good to merge.

PR #383, opened 2020-07-10
Added ROS2StateProbe class (fixes #382)

### LINKED ISSUE #382: Implement ROS2StateProbe

mvkashyap:


src/roswire/proxy/ros2state.py
```diff
@@ -0,0 +1,60 @@
+# -*- coding: utf-8 -*-
+__all__ = ('SystemState', 'ROS2SystemStateProbe')
```
ChrisTimperley, line 2: No need to export `SystemState` here.

src/roswire/proxy/ros2state.py
```diff
@@ -0,0 +1,60 @@
...
+                    elif "Action Servers" in line:
+                        mode = 'None'
+                    elif mode != 'None':
+                        topic, space, fmt = line.partition(':')
+                        if topic in mode_dict[mode]:
+                            (mode_dict[mode])[topic].append(name)
+                        else:
+                            (mode_dict[mode])[topic] = [name]
+            except dockerblade.exceptions.CalledProcessError as error:
+                raise error
```
ChrisTimperley, line 53: In the general case, if there's no specific handling of the exception, the `try` and `except` can be dropped and the program will be equivalent (and easier to understand). In this case, you probably want to add a logging call so that we know where an error has occurred.

src/roswire/proxy/ros2state.py
```diff
@@ -0,0 +1,60 @@
...
+        try:
+            output = self._app_instance.shell.check_output(command, text=True)
+        except dockerblade.exceptions.CalledProcessError as error:
+            raise error
+        node_names = output.split('\r\n')
+        for name in node_names:
+            info = f"ros2 node info '{name}'"
+            mode = 'None'
+            try:
+                output = self._app_instance.shell.check_output(info, text=True)
```
ChrisTimperley, line 34: The `except` clause should come immediately after the `output = ...` line. In general, try to keep the scope of the `try` block as small as possible. In this case, the error can only occur on the `output = ...` line.

src/roswire/proxy/ros2state.py
```diff
@@ -0,0 +1,60 @@
...
+                for line in lines:
+                    if "Publications:" in line:
+                        mode = 'pub'
+                    elif "Subscriptions:" in line:
+                        mode = 'sub'
+                    elif "Services" in line:
+                        mode = 'serv'
+                    elif "Action Servers" in line:
+                        mode = 'None'
+                    elif mode != 'None':
```
ChrisTimperley, line 46: This looks like a bug. I think you mean `if` rather than `elif`. (You may also want to add a blank line above to indicate that this isn't part of the `if/elif` block above.)

ChrisTimperley:
It looks like another `.swp` file has accidentally been added to the repo.

src/roswire/ros2/state.py
```diff
@@ -14,12 +18,53 @@
...
         """Obtains the instantaneous state of the associated ROS system."""
-        raise NotImplementedError
+        mode_dict: Dict[str, Dict[str, List[str]]] = {'pub': {},
+                                                      'sub': {},
+                                                      'serv': {}}
+        command = "rosnode list"
+        try:
+            output = self._app_instance.shell.check_output(command, text=True)
+        except dockerblade.exceptions.CalledProcessError:
+            logger.debug("Unable to retrieve rosnode list from command line")
```
ChrisTimperley, line 36: This `except` block is missing a `raise`. Right now, the program will write to the log and then continue executing the rest of the method.

src/roswire/ros2/state.py
```diff
@@ -14,12 +18,55 @@
...
-        raise NotImplementedError
+        node_to_state: Dict[Optional[str], Dict[str, List[str]]] = {'pub': {},
+                                                                    'sub': {},
+                                                                    'serv': {}}
+        command = "rosnode list"
+        try:
+            output = self._app_instance.shell.check_output(command, text=True)
+        except dockerblade.exceptions.CalledProcessError as error:
+            logger.debug("Unable to retrieve rosnode list from command line")
+            raise error
```
ChrisTimperley, line 37: Incredibly minor improvement: There's actually no need to include `error` in `raise error` here. Since we don't actually inspect or manipulate `error`, we can do the following:
```
except SomeException:
    # do some unrelated things
    raise
```

src/roswire/ros2/state.py
```diff
@@ -14,12 +18,55 @@
...
+        node_names = output.split('\r\n')
+        for node_name in node_names:
+            info = f"ros2 node info '{node_name}'"
```
ChrisTimperley, line 40: Let's rename this to `command_info` to make it clear that this is a command (rather than the output from the command).

src/roswire/ros2/state.py
```diff
@@ -14,12 +18,55 @@
...
+            mode: Optional[str] = None
+            try:
+                output = self._app_instance.shell.check_output(info, text=True)
```
ChrisTimperley, line 43: Writing `self._app_instance.shell.` multiple times can get a bit tedious. Instead, you can create a new local variable `shell` at the top of the method and essentially use that as an alias:
```
shell = elf._app_instance.shell
```

src/roswire/ros2/state.py
```diff
@@ -14,12 +18,55 @@
...
+            except dockerblade.exceptions.CalledProcessError as error:
+                logger.debug(f"Unable to retrieve {info}")
+                raise error
+            output = output.replace(' ', '')
+            lines = output.split('\r\n')
+            for line in lines:
+                if "Publications:" in line:
+                    mode = 'pub'
+                    continue
```
ChrisTimperley, line 52: This will skip ahead to the next `line` in `lines`. I don't think that's the desired behaviour here. If you want to double check and improve the quality of the codebase, you could write a test case to test this class. (I can give you some pointers on how to do that via Slack, if you want.)

src/roswire/ros2/node.py
```diff
@@ -10,7 +10,7 @@
 
 
 @attr.s(frozen=True, auto_attribs=True, slots=True)
-class ROS2Node:
+class ROS2Node():
```
ChrisTimperley, line 13: No need for `()`

src/roswire/ros2/state.py
```diff
@@ -1,8 +1,12 @@
 # -*- coding: utf-8 -*-
 __all__ = ('ROS2StateProbe',)
 
+import dockerblade
```
ChrisTimperley, line 4: Apologies for a very nitpicky comment. Imports should be grouped into three blocks: 1. standard library imports, 2. third-party imports, 3. relative imports. (See: https://www.python.org/dev/peps/pep-0008/#imports)

mvkashyap: Oh I didn't know this. Thanks for telling me! I'll fix it for this and everything else

ChrisTimperley:
It looks like a few temporary and miscellaneous files have sneaked into this PR.

src/roswire/ros2/ros2.py
```diff
@@ -38,7 +38,9 @@ def __attrs_post_init__(self) -> None:
     def for_app_instance(cls, app_instance: 'AppInstance') -> 'ROS2':
         return ROS2(app_instance=app_instance)
 
-    # TODO add launch manager
+    @property
+    def launch_manager(self) -> ROS2LaunchManager:
```
ChrisTimperley, line 42: No need for this property.

test/test_state.py
```diff
@@ -0,0 +1,67 @@
+# -*- coding: utf-8 -*-
+import pytest
+
+import time
+
+import dockerblade
+
+@pytest.mark.parametrize('sut', ['turtlebot3-ros2'], indirect=True)
+def test_state_publishers(sut):
+    sut.ros2.launch_manager.launch('simple.launch.py', package='launch')
```
ChrisTimperley, line 10: `launch` is the name of a package?
Also, this call should be:
```
sut.ros2.launch(....)
```
(we don't need the extra stuff :-))

test/test_state.py
```diff
@@ -0,0 +1,67 @@
...
+    actual_state = sut.ros2.state
+    actual_publishers = {pub for pub in actual_state.publishers}
```
ChrisTimperley, line 12: slightly neater: `set(actual_state.publishers)`

src/roswire/ros2/state.py
ChrisTimperley: There's still at least one bug to be fixed here. Note that `continue` will immediately skip to the next iteration in the loop, and so if any of the if/elif conditions between 53 and 62 are satisfied, that particular line in the file will be ignored.

If the current tests don't reveal the bug, you'll want to refine / add to them to expose this particular bug.

ChrisTimperley: Ah, my bad. It looks like this is switching the mode that's used to parse the subsequent lines. It's probably worth adding a comment to the top of this `if/elif` block to explain the logic.

src/roswire/ros2/state.py
```diff
@@ -14,12 +18,60 @@
...
+                    continue
+                elif "Action Servers:" in line:
+                    break
+
+                if mode:
+                    name = line.partition(':')[0]
+                    if name in node_to_state[mode]:
+                        node_to_state[mode][name].append(node_name)
+                    else:
+                        node_to_state[mode][name] = [node_name]
```
ChrisTimperley, line 70: Should this be a set rather than a list?

src/roswire/definitions/msg.py
```diff
@@ -17,7 +17,7 @@
 from .decode import is_simple
 from .. import exceptions as exc
 
-R_TYPE = r"[a-zA-Z0-9_/]+(?:\[\d*\])?"
+R_TYPE = r"[a-zA-Z0-9_/]+(?:\[(?:<=)?\d*\])?"
```
ChrisTimperley, line 20: It might be worth just separating the fixes to this file into another PR since they're not really related to the state probe directly.

src/roswire/definitions/package.py
```diff
@@ -100,18 +100,29 @@ def paths(shell: dockerblade.Shell,
...
+            package_names = '/opt/ros/dashing/share/' \
+                            + 'ament_index/resource_index' \
+                            + '/packages'
+            all_packages = files.listdir(package_names)
+            package_dirs = []
+            for p in all_packages:
+                if files.exists('/opt/ros/dashing/share/' + p):
+                    package_dirs.append('/ros_ws/src' + p)
             paths.extend(package_dirs)
+        else:
```
ChrisTimperley, line 114: It looks like you probably want to add `_paths_ros1` and `_paths_ros2` method. You can then look at the `ROS_DISTRO` environment variable to determine if ROS1 or ROS2 is being used.

src/roswire/definitions/package.py
```diff
@@ -100,18 +100,29 @@ def paths(shell: dockerblade.Shell,
...
         paths: List[str] = []
-        for path in package_paths:
-            try:
-                all_packages = files.find(path, 'package.xml')
-            except dockerblade.exceptions.DockerBladeException:
-                logger.warning('unable to find directory in ROS_PACKAGE_PATH:'
-                               f' {path}')
-                continue
-            package_dirs = [os.path.dirname(p) for p in all_packages]
+        if files.exists('/opt/ros/dashing/'):
```
ChrisTimperley, line 104: There's a couple of things to improve here:
(1) this is hardcoded for our particular Docker images, and only for Dashing-based ROS2 images (i.e., it won't work for Foxy or Eloquent). (2) this appears to ignore any and all of the colon workspaces (e.g., `/ros_ws/src`).

src/roswire/definitions/package.py
```diff
@@ -100,18 +100,29 @@ def paths(shell: dockerblade.Shell,
...
+            package_names = '/opt/ros/dashing/share/' \
+                            + 'ament_index/resource_index' \
+                            + '/packages'
+            all_packages = files.listdir(package_names)
+            package_dirs = []
+            for p in all_packages:
+                if files.exists('/opt/ros/dashing/share/' + p):
```
ChrisTimperley, line 111: For this particular implementation, this if condition is a tautological. Each `p` comes from `listdir`, so we know that the file definitely exists. (If there's a case where this isn't true, then there's a nasty bug lurking.)

src/roswire/definitions/msg.py
mvkashyap: I can't test anything related to ROS2 unless I have this package modification, which is why I'm including it in all of the PRs

src/roswire/definitions/package.py
mvkashyap: I just created a new PR for this, if that's okay.

src/roswire/ros2/node_manager.py
```diff
@@ -17,7 +17,7 @@
 @attr.s(frozen=True, auto_attribs=True, slots=True)
 class ROS2NodeManager(Mapping[str, ROS2Node]):
     """Provides an interface for interacting with ROS2 nodes."""
-    app_instance: 'AppInstance'
+    app_instance: 'AppInstance' = attr.ib()
```
ChrisTimperley, line 20: Just a heads up: `attr.ib()` isn't actually needed here since we use `auto_attribs=True` at the top. Using either approach is fine, but if we're going to use `attr.ib()`, we probably want to remove the `auto_attribs=True` keyword to some unexpected accidents involving class vars.

src/roswire/definitions/msg.py
ChrisTimperley: Fair enough. Situations like these are always annoying to deal with (i.e., when there's some common building block that multiple PRs need). For the future, it can help to indicate in the PR description that merging a particular PR (e.g., this one) is blocked on the merging of another (e.g., the package database PR).

PR #388, opened 2020-07-20
Added partial implementation for ROS2Node and ROS2NodeManager

mvkashyap:


src/roswire/ros2/node.py
```diff
@@ -5,20 +5,34 @@
...
     """Provides an interface for interacting with a ROS2 node.
 
     Attributes
     ----------
     name: str
         The name of the node.
     """
-    app_instance: 'AppInstance'
+    app_instance: 'AppInstance' = attr.ib()
+    __name: str
```
ChrisTimperley, line 24: `__name` doesn't do anything here. You probably want to use `attr.ib()` to create a public `name` attribute. Notice that `frozen=True` appears in the `@attr.s` decorator for the class: That instructs attrs to create an immutable class. That is, users can read the `name` attribute but they can't write to it.

src/roswire/ros2/node.py
```diff
@@ -5,20 +5,34 @@
...
+
+    @classmethod
+    def for_app_instance_and_name(cls,
+                                  name: str,
```
ChrisTimperley, line 28: I would swap `name` and `app_instance` to match the order specified in the method name.

src/roswire/ros2/node.py
```diff
@@ -5,20 +5,34 @@
...
+                                  app_instance: 'AppInstance'
+                                  ) -> 'ROS2Node':
+        return ROS2Node(name=name, app_instance=app_instance)
+
+    @property
+    def name(self) -> str:
+        return self.__name
 
     def is_alive(self) -> bool:
         """Determines whether this node is alive."""
```
ChrisTimperley, line 38: You could conceivably implement `is_alive` and `shutdown`.

src/roswire/ros2/node_manager.py
```diff
@@ -1,23 +1,26 @@
...
     from ..app import AppInstance
 
 
 @attr.s(frozen=True, auto_attribs=True, slots=True)
-class ROS2NodeManager(Mapping[str, ROS2Node]):
+class ROS2NodeManager(NodeManager):
     """Provides an interface for interacting with ROS2 nodes."""
-    app_instance: 'AppInstance'
+    app_instance: 'AppInstance' = attr.ib()
+    state: 'ROS2StateProbe' = ROS2StateProbe.for_app_instance(app_instance)
```
ChrisTimperley, line 23: `state` doesn't achieve what you're after here (and will likely lead to a crash at runtime). In its current form, this code tries to create a class variable `state` using an uninstantiated instance variable `app_instance`.

Implementing this requires a more advanced use of attrs. To implement the `NodeManager` interface, you do indeed want to have a `state` attribute, but you want to instantiate the state probe in the constructor. However, attrs doesn't have a conventional `__init__` constructor! Fortunately, attrs calls a `__attrs_post_init__` method after it has initialised an object but before that object is given to any potential caller.

The trick here is to declare the `state` attribute using `attr.ib(init=False)`, and to defer its initialisation to `__attrs_post_init__`. Normally, you would just be able to do `self.state = ROS2StateProbe.for_app_instance(app_instance)` inside `__attrs_post_init__`. However, here we have made `ROS2NodeManager` a frozen (i.e., immutable) class. That means we can't directly modify `self`! Thankfully, we can use a trick to set the value of the `state` attribute inside `__attrs_post_init__`: https://github.com/ChrisTimperley/roswire/blob/9e7cffa35693d40bdf31a8ddb757b8ef1b8f6c57/src/roswire/ros2/ros2.py#L27-L35

ChrisTimperley: Having said all of this, I realise that `state` is not a `ROS2StateProbe` in the `NodeManager` interface. `state` should be implemented as a `@property` that uses the state probe behind the scenes to obtain the state of the system.

In that case, the probe should become `_state_probe`, but all of the other comments above still apply.

src/roswire/ros2/node_manager.py
```diff
@@ -43,15 +46,15 @@ def __getitem__(self, name: str) -> ROS2Node:
         NodeNotFoundError
             If there is no node with the given name.
         """
-        raise NotImplementedError
+        return ROS2Node.for_app_instance_and_name(name, self.app_instance)
 
     def __len__(self) -> int:
         """Returns a count of the number of active nodes."""
-        raise NotImplementedError
+        return len(self.state.probe().nodes)
```
ChrisTimperley, line 53: You'll be able to use `self.state.nodes` here instead.

src/roswire/ros2/node.py
```diff
@@ -4,26 +4,50 @@
...
     """Provides an interface for interacting with a ROS2 node.
 
     Attributes
     ----------
     name: str
         The name of the node.
     """
-    app_instance: 'AppInstance'
+    app_instance: 'AppInstance' = attr.ib()
+    node_name: str = attr.ib()
```
ChrisTimperley, line 27: We can just use `name` here rather than `node_name`. We don't need the extra `@property` :slightly_smiling_face:

src/roswire/ros2/node.py
```diff
@@ -4,26 +4,50 @@
...
+        return False
 
     def shutdown(self) -> None:
         """Instructs this node to shutdown."""
-        raise NotImplementedError
+        command = f"ros2 lifecycle set {self.node_name} shutdown"
+        try:
+            self.app_instance.shell.run(command)
+        except dockerblade.exceptions.CalledProcessError:
+            logger.debug(f"Unable to shutdown node {self.node_name}")
```
ChrisTimperley, line 53: This should probably raise some kind of exception if the node can't be shutdown. We may want to add a new exception for that.

src/roswire/ros2/node_manager.py
```diff
@@ -1,30 +1,38 @@
...
     from ..app import AppInstance
 
 
 @attr.s(frozen=True, auto_attribs=True, slots=True)
-class ROS2NodeManager(Mapping[str, ROS2Node]):
+class ROS2NodeManager(NodeManager):
     """Provides an interface for interacting with ROS2 nodes."""
-    app_instance: 'AppInstance'
+    app_instance: 'AppInstance' = attr.ib()
+    _state_probe: 'ROS2StateProbe' = attr.ib(init=False)
```
ChrisTimperley, line 24: Adding a comment here just to make sure that we remember to include an `__attrs_post_init__` method before merging.

src/roswire/ros2/node.py
```diff
@@ -4,26 +4,50 @@
...
 
     Attributes
     ----------
     name: str
         The name of the node.
     """
-    app_instance: 'AppInstance'
+    app_instance: 'AppInstance' = attr.ib()
+    node_name: str = attr.ib()
+    _state_probe: 'ROS2StateProbe' = attr.ib(init=False)
```
ChrisTimperley, line 28: Adding a comment here just to make sure that we remember to include an `__attrs_post_init__` method before merging.

src/roswire/ros2/node.py
mvkashyap: I could add an exception for that called `NodeShutdownError` or something like that

ChrisTimperley: That sounds great. We might even want to refine the general interface of `shutdown` in `Node` to take an optional keyword argument `ignore_errors` with a default value of `False`. If `ignore_errors` is `True`, we don't raise an exception; if it's `False`, we go ahead and raise a `NodeShutdownError`.

src/roswire/ros2/node.py
```diff
@@ -2,28 +2,48 @@
...
+        return False
 
     def shutdown(self) -> None:
         """Instructs this node to shutdown."""
-        raise NotImplementedError
+        command = f"ros2 lifecycle set {self.name} shutdown"
+        try:
+            self.app_instance.shell.run(command)
+        except dockerblade.exceptions.CalledProcessError:
+            logger.debug(f"Unable to shutdown node {self.name}")
```
ChrisTimperley, line 49: Remember to add `NodeShutdownError`

src/roswire/exceptions.py
```diff
@@ -153,3 +153,10 @@ class SourceNotFoundError(ValueError, ROSWireException):
 
     def __str__(self) -> str:
         return f"source not found in container: {self.source}"
+
+
+class NodeShutdownError(ROSWireException):
+    """ROS2 node could not be killed"""
+    def __init__(self, node: str):
```
ChrisTimperley, line 160: Don't forget to add `-> None` to the type definition for `__init__`.

src/roswire/ros2/node.py
```diff
@@ -40,10 +41,12 @@ def is_alive(self) -> bool:
             return True
         return False
 
-    def shutdown(self) -> None:
+    def shutdown(self, ignore_errors: bool = False) -> None:
         """Instructs this node to shutdown."""
```
ChrisTimperley, line 45: We should add a "Raises" section to this docstring that states this method produces a `NodeShutdownError` in the event of an error.

src/roswire/ros2/node_manager.py
```diff
@@ -23,6 +23,10 @@ class ROS2NodeManager(NodeManager):
     app_instance: 'AppInstance' = attr.ib()
     _state_probe: 'ROS2StateProbe' = attr.ib(init=False)
 
+    def __attrs_post_init__(self) -> None:
+        app_instance = self.app_instance
+        state_probe = self._state_probe
```
ChrisTimperley, line 28: `self._state_probe` hasn't been defined, so this line will throw an exception. This method should be used to define `self._state_probe`. If the object wasn't frozen, then you could do `self._state_probe = ...`, but since the object is frozen, you'll need to use `object.__setattr__` (see https://github.com/ChrisTimperley/roswire/blob/9e7cffa35693d40bdf31a8ddb757b8ef1b8f6c57/src/roswire/ros2/ros2.py#L32)

src/roswire/ros2/node.py
```diff
@@ -2,28 +2,57 @@
 __all__ = ('ROS2Node',)
 
 import typing
-
 import attr
 
+import logger
```
ChrisTimperley, line 7:
```
from loguru import logger
````

File: src/roswire/ros2/node.py [hunk 15]
```diff
@@ -2,28 +2,57 @@
 __all__ = ('ROS2Node',)
 
 import typing
-
 import attr
```
ChrisTimperley, line 5: Should appear in same block of imports as `dockerblade`. The `from` import can go above this one.

src/roswire/ros2/node.py
ChrisTimperley: Reminder about this

src/roswire/ros2/node_manager.py
```diff
@@ -1,30 +1,42 @@
...
-class ROS2NodeManager(Mapping[str, ROS2Node]):
+class ROS2NodeManager(NodeManager):
     """Provides an interface for interacting with ROS2 nodes."""
-    app_instance: 'AppInstance'
+    app_instance: 'AppInstance' = attr.ib()
+    _state_probe: 'ROS2StateProbe' = attr.ib(init=False)
+
+    def __attrs_post_init__(self) -> None:
+        _state_probe = ROS2StateProbe.for_app_instance(self.app_instance)
+        object.__setattr__(self, 'state_probe', _state_probe)
```
ChrisTimperley, line 28: The second argument, `'state_probe'`, gives the name of the attribute that should be set as a string. In this case, we want to set the `_state_probe` attribute, so we should add the `_` prefix to the string. Note that the local variable `_state_probe` does not need a leading underscore (it can't be accessed outside of the scope of this method).

src/roswire/ros2/node.py
```diff
@@ -1,29 +1,62 @@
 # -*- coding: utf-8 -*-
 __all__ = ('ROS2Node',)
 
-import typing
+from loguru import logger
```
ChrisTimperley, line 4: Sorry to be super pedantic: `loguru` and `dockerblade` should go together in a second block of imports. (See comment below.)

src/roswire/ros2/node.py
ChrisTimperley: `typing` is part of the standard library, so the new line should remain.

src/roswire/ros2/node.py
```diff
@@ -2,28 +2,61 @@
...
+class ROS2Node(Node):
     """Provides an interface for interacting with a ROS2 node.
 
     Attributes
     ----------
     name: str
         The name of the node.
     """
-    app_instance: 'AppInstance'
+    app_instance: 'AppInstance' = attr.ib()
```
ChrisTimperley, line 27: Since `auto_attribs` is being used, we don't need to explicitly state `= attr.ib()` for `app_instance` or `name`.

src/roswire/ros2/node_manager.py
```diff
@@ -1,30 +1,42 @@
 # -*- coding: utf-8 -*-
 __all__ = ('ROS2NodeManager',)
 
-from typing import Iterator, Mapping
+from typing import Iterator
 import typing
+import attr
```
ChrisTimperley, line 6: `attr` is a third-party module, so it should be grouped with `loguru`. (See previous version.)

PR #389, opened 2020-07-21
Added deep_clean method to CatkinInference and implemented CatkinMake (fixes #221)

### LINKED ISSUE #221: Add proxy for catkin_make

afsafzal:


afsafzal:
@ChrisTimperley I added the `deep_clean` function to `CatkinInterface`. With this functionality I no longer need `CatkinMake` and I can simply just use `CatkinTools`, so if you think the repeated functionality is code smell I can return both `build` and `clean` back to `raise NotImplementedError` as before.

src/roswire/proxy/catkin.py
```diff
@@ -59,12 +73,34 @@ def build(self,
         """
         ...
 
+    def deep_clean(self) -> None:
+        """Removes the build, devel, and install from the workspace.
+
+        Raises
+        ------
+        CatkinException:
```
ChrisTimperley, line 81: no need for trailing `:`

src/roswire/proxy/catkin.py
```diff
@@ -59,12 +73,34 @@ def build(self,
         """
         ...
 
+    def deep_clean(self) -> None:
+        """Removes the build, devel, and install from the workspace.
```
ChrisTimperley, line 77: `and install [directories]`

src/roswire/proxy/catkin.py
```diff
@@ -135,13 +171,33 @@ class CatkinMake(CatkinInterface):
...
+            command += ['--pkg']
+            command += [shlex.quote(p) for p in packages]
+        if not context:
+            context = self.directory
+
+        command_str = ' '.join(command)
+        logger.debug(f"cleaning via: {command_str}")
+        result = shell.run(command_str, cwd=context, text=True)
+        duration_mins = result.duration / 60
+        logger.debug("clean completed after %.2f minutes [retcode: %d]:\n%s",
```
ChrisTimperley, line 195: `%s`, `%.2f`, and so on won't work by default in loguru. We should use f-strings instead:
`clean completed after {durations_mins:.2f} minutes ...`. (Note that you can break Python strings over multiple lines.)

ChrisTimperley:
Overall, the changes look nice. Just a few minor tweaks and then we're good to merge.

ChrisTimperley:
We probably want to update the CHANGELOG here, too.

PR #395, opened 2020-08-12
Added tests for ROS2 launch

mvkashyap:


src/roswire/ros2/ros2.py
```diff
@@ -21,25 +19,16 @@ class ROS2:
     app_instance: 'AppInstance'
     nodes: ROS2NodeManager = attr.ib(init=False)
     services: ROS2ServiceManager = attr.ib(init=False)
-    _state_probe: ROS2StateProbe = attr.ib(init=False)
```
ChrisTimperley, line 24: Why remove the state probe?

mvkashyap: `StateProbe.for_app_instance()` is not implemented in this branch, so it was giving me an error when I was trying to run test cases for `launch` and `locate`.

ChrisTimperley: It still looks like state probe has been removed.

src/roswire/ros2/ros2.py
```diff
@@ -21,13 +20,12 @@ class ROS2:
...
     launch: ROS2LaunchManager = attr.ib(init=False)
 
     def __attrs_post_init__(self) -> None:
         nodes = ROS2NodeManager.for_app_instance(self.app_instance)
         services = ROS2ServiceManager.for_app_instance(self.app_instance)
-        state_probe = ROS2StateProbe.for_app_instance(self.app_instance)
+        state_probe: ROS2StateProbe = attr.ib(init=False)
```
ChrisTimperley, line 28: The old version of this line was correct.

src/roswire/ros2/ros2.py
```diff
@@ -37,7 +35,3 @@ def __attrs_post_init__(self) -> None:
     @classmethod
     def for_app_instance(cls, app_instance: 'AppInstance') -> 'ROS2':
         return ROS2(app_instance=app_instance)
-
-    @property
-    def state(self) -> SystemState:
```
ChrisTimperley, line 42: This should be restored.

src/roswire/ros2/ros2.py
```diff
@@ -8,7 +8,6 @@
 from .node_manager import ROS2NodeManager
 from .service_manager import ROS2ServiceManager
 from .state import ROS2StateProbe
-from ..proxy import SystemState
```
ChrisTimperley, line 11: This should be restored.

src/roswire/ros2/launch.py
```diff
@@ -167,17 +174,20 @@ def launch(self,
...
             launch_prefixes = {}
-        filename = self.locate(filename, package=package)
 
         if node_to_remappings or launch_prefixes:
             m = "Requires self.read: not yet implemented"
             raise NotImplementedError(m)
-
-        cmd = ['ros2 launch', shlex.quote(filename)]
+        if package:
+            cmd = ['ros2 launch', (package + ' ' + filename_without_path)]
```
ChrisTimperley, line 187: Use `shlex.quote` to protect against malformed inputs. Also, we join the string using `' '` later, so there's no need for `(foo + ' ' + bar)`. Prefer:

`['ros2 launch', shlex.quote(package), shlex.quote(filename_without_path)]`

ChrisTimperley: `filename_without_path` should be moved to inside this `if` block.

src/roswire/ros2/launch.py
```diff
@@ -111,17 +112,23 @@ def locate(self, filename: str, *, package: Optional[str] = None) -> str:
...
-        package_path = app_description.packages[package].path
-        filename = os.path.join(package_path, 'launch', filename_original)
-        if not self._app_instance.files.isfile(filename):
-            raise exc.LaunchFileNotFound(path=filename)
-        logger.debug('determined location of launch file'
-                     f' [{filename_original}] in package [{package}]: '
-                     f'{filename}')
-        return filename
+        else:
+            command = 'ros2 package prefix ' + shlex.quote(package)
```
ChrisTimperley, line 116: We should retrieve the package location from the package database to avoid a redundant and very expensive API call.

test/test_roslaunch.py
```diff
@@ -62,3 +62,31 @@ def test_remappings(sut):
...
+def test_locate_with_package_ros2(sut):
+    actual_path = sut.ros2.launch.locate('cartographer.launch.py', package='turtlebot3_cartographer')
+    expected_path = '/ros_ws/src/turtlebot3/turtlebot3_cartographer/launch/cartographer.launch.py'
+    assert actual_path == expected_path
+
+
+@pytest.mark.parametrize('sut', ['turtlebot3-ros2'], indirect=True)
+def test_locate_without_package_ros2(sut):
+    actual_path = sut.ros2.launch.locate('/ros_ws/src/turtlebot3/turtlebot3_cartographer/launch/cartographer.launch.py')
+    expected_path = expected_path = '/ros_ws/src/turtlebot3/turtlebot3_cartographer/launch/cartographer.launch.py'
```
ChrisTimperley, line 76: `expected_path = expected_path = ` looks like a typo?

src/roswire/ros2/launch.py
mvkashyap: I'm sorry, I'm not exactly sure how to do this. How do I access the package database?

ChrisTimperley: No worries. `self._app_instance` gives you access to the `AppInstance` to which the launch manager belongs. From there, you can get a description of the associated app:
```
app_description = self._app_instance.app.description
```

You can then get access to a description of the package from the `PackageDatabase`:
```
package_description = app_description.packages[package]
```

and finally:
```
package_location = package_description.path
```

PR #396, opened 2020-08-13
ROS2 package database

mvkashyap:


ChrisTimperley:
It looks like there's three accidentally added files: `app.txt`, `test.txt`, and a `*.swp` file.

src/roswire/definitions/package.py
```diff
@@ -114,6 +114,32 @@ def paths(shell: dockerblade.Shell,
...
+        paths.extend(package_dirs)
+        return paths
+
+    @staticmethod
+    def paths(shell: dockerblade.Shell,
+              files: dockerblade.FileSystem
+              ) -> List[str]:
+        """Parses :code:`ROS_PACKAGE_PATH` for a given shell."""
+        distro = shell.environ('ROS_DISTRO')
+        if distro == 'foxy' or distro == 'eloquent' or distro == 'dashing':
```
ChrisTimperley, line 139:
```
ROS2_DISTROS = ('dashing', 'eloquent', 'foxy')
if distro in ROS2_DISTROS:
```

src/roswire/definitions/package.py
```diff
@@ -114,6 +114,32 @@ def paths(shell: dockerblade.Shell,
...
+                        '/ament_index/resource_index/packages'
+        all_packages = files.listdir(package_names)
+        package_dirs = []
+        for p in all_packages:
+            package_dirs.append('/opt/ros/' + distro + '/share/' + p)
+        paths.extend(package_dirs)
+        return paths
+
+    @staticmethod
+    def paths(shell: dockerblade.Shell,
```
ChrisTimperley, line 134: Prefer @classmethod

src/roswire/definitions/package.py
```diff
@@ -114,6 +114,32 @@ def paths(shell: dockerblade.Shell,
...
+              files: dockerblade.FileSystem
+              ) -> List[str]:
+        """Parses :code:`ROS_PACKAGE_PATH` for a given shell."""
+        distro = shell.environ('ROS_DISTRO')
+        if distro == 'foxy' or distro == 'eloquent' or distro == 'dashing':
+            return PackageDatabase._paths_ros2(shell, files)
```
ChrisTimperley, line 140: Prefer
```
return cls._path_ros2(....)
```

src/roswire/definitions/package.py
```diff
@@ -114,6 +114,32 @@ def paths(shell: dockerblade.Shell,
...
+        return PackageDatabase._paths_ros1(shell, files)
```
ChrisTimperley, line 141: Same as above

src/roswire/ros2/node.py
```diff
@@ -2,28 +2,39 @@
...
+from ..interface import Node
+from .state import ROS2StateProbe
+
 if typing.TYPE_CHECKING:
     from ..app import AppInstance
 
 
 @attr.s(frozen=True, auto_attribs=True, slots=True)
-class ROS2Node:
+class ROS2Node(Node):
```
ChrisTimperley, line 15: The changes here and in node_manager.py don't seem to be related to the topic of the PR. The right call here is to merge the rest of the changes, and then submit these node-related changes in a separate PR.

src/roswire/ros2/node_manager.py
```diff
@@ -1,30 +1,38 @@
 # -*- coding: utf-8 -*-
 __all__ = ('ROS2NodeManager',)
 
-from typing import Iterator, Mapping
+from typing import Iterator
 import typing
+import attr
```
ChrisTimperley, line 6: Bit of a nitpick: standard library imports should be separated from external imports (see https://www.python.org/dev/peps/pep-0008/#imports)

src/roswire/ros2/node.py
mvkashyap: This was actually an accident--I forgot to switch the branches when I started working on the ROS2NodeManager branch.

src/roswire/definitions/package.py
```diff
@@ -114,6 +114,34 @@ def paths(shell: dockerblade.Shell,
...
         return paths
 
+    @staticmethod
+    def _paths_ros2(shell: dockerblade.Shell,
+                    files: dockerblade.FileSystem
+                    ) -> List[str]:
+        """Returns paths of packages"""
+        paths: List[str] = []
+        distro = shell.environ('ROS_DISTRO')
+        package_names = '/opt/ros/' + distro + '/share' \
```
ChrisTimperley, line 124: We're almost there. Just a couple of issues to deal with:
(1) ROS is assumed to be installed to a specific location, and, more importantly, (2) packages within the colcon workspace (e.g., `/ros_ws`) are ignored.

src/roswire/definitions/package.py
```diff
@@ -114,6 +114,43 @@ def paths(shell: dockerblade.Shell,
...
+        all_packages = package_str.split('\r\n')
+        package_dirs: List[str] = []
+        for p in all_packages:
+            command = 'ros2 pkg prefix ' + p
+            try:
+                package_path = shell.check_output(command, text=True)
+            except dockerblade.exceptions.CalledProcessError:
+                logger.debug(f'unable to find package {p}')
+                raise
+            package_dirs.append(package_path)
```
ChrisTimperley, line 138: This is a subtle bug. `package_dirs` isn't clearly between loop iterations, and so each time `.extend` is called on the line below, every existing path is reinserted into `paths`. The fix here is to just drop `package_dirs` entirely and append to `paths`.

test/test_package.py
```diff
@@ -253,3 +253,186 @@ def test_database_from_paths(filesystem):
     db = PackageDatabase.from_paths(filesystem, paths)
     assert len(db) == len(paths)
     assert set(db) == {'angles', 'tf2', 'tf2_msgs', 'tf2_py', 'tf2_ros'}
+
+
+@pytest.mark.parametrize('sut', ['turtlebot3-ros2'], indirect=True)
+def test_package_location_ros2(sut):
+    expected_paths = {'/ros_ws/install/pcl_conversions',
```
ChrisTimperley, line 260: Does this include `/opt/ros` packages?

src/roswire/definitions/package.py
```diff
@@ -114,6 +114,43 @@ def paths(shell: dockerblade.Shell,
...
+        command = "ros2 pkg list"
+        try:
+            package_str = shell.check_output(command, text=True)
+        except dockerblade.exceptions.CalledProcessError:
+            logger.debug(f'unable to find packages using {command}')
+            raise
+        all_packages = package_str.split('\r\n')
+        package_dirs: List[str] = []
+        for p in all_packages:
+            command = 'ros2 pkg prefix ' + p
```
ChrisTimperley, line 132: To avoid surprises, use `shlex.quote(p)`.

src/roswire/definitions/package.py
```diff
@@ -114,6 +114,43 @@ def paths(shell: dockerblade.Shell,
...
 
+    @staticmethod
+    def _paths_ros2(shell: dockerblade.Shell,
+                    files: dockerblade.FileSystem
+                    ) -> List[str]:
+        """Returns paths of packages"""
+        paths: List[str] = []
+        command = "ros2 pkg list"
+        try:
+            package_str = shell.check_output(command, text=True)
```
ChrisTimperley, line 125: `ros2 pkg list` can just be used directly here.

test/test_package.py
mvkashyap: It should. The `/ros_ws/install` directory includes files the packages in both `/opt/ros` and the `catkin workspace`

PR #399, opened 2020-09-18
Added ROSDistribution and ROSVersion information to AppDescription (fixes #398)

### LINKED ISSUE #398: Add ros_distro to AppDescription and add ROSDistro enum

ChrisTimperley:


test/apps/fetch.yml
```diff
@@ -1,4 +1,5 @@
 image: therobotcooperative/fetch
+distribution: melodic
```
schmerl, line 2: If you're getting the ROS version from the ROS_DISTRO environment in the container, does the information need to be duplicated here?

ChrisTimperley: This is to remember the ROS distribution associated with a given image so that we can make certain decisions without needing to launch a container.

PR #405, opened 2020-09-29
Added __ne__ and __lt__ to ROSDistribution

ChrisTimperley:


src/roswire/distribution.py
```diff
@@ -2,7 +2,7 @@
 __all__ = ('ROSDistribution', 'ROSVersion')
 
 import enum
-from typing import Sequence
+from typing import Any, Sequence
 
 
 class ROSVersion(enum.IntEnum):
```
schmerl, line 8: Don't you need @total_ordering annotation?

ChrisTimperley: Yep :+1:

schmerl (approved):

PR #424, opened 2020-10-19
Fixed some relative imports

schmerl:
Some relative imports inside `if typing.TYPECHECKING` aren't checked by MyPy and were wrong after the move.


ChrisTimperley (approved):
Nice catch! 👍🏻

PR #426, opened 2020-10-20
Moved ROS1 from proxy submodule into ros1 submodule

schmerl:


src/roswire/proxy/__init__.py
```diff
@@ -1,2 +1 @@
 # -*- coding: utf-8 -*-
```
ChrisTimperley, line 1: We can remove the `proxy` directory entirely.

src/roswire/ros1/__init__.py
```diff
@@ -1,7 +1,2 @@
 # -*- coding: utf-8 -*-
-from .node import ROS1Node
```
ChrisTimperley, line 2: These lines shouldn't be removed.

src/roswire/ros1/ros1.py
```diff
@@ -74,7 +77,7 @@ def __init__(self,
                            self.__shell)
         self.__state_probe: SystemStateProbe = \
             SystemStateProbe.via_xmlrpc_connection(self.__connection)
-        self.roslaunch: ROSLaunchManager = \
+        self.roslaunch: 'ROSLaunchManager' = \
```
ChrisTimperley, line 80: No need for quotes.

test/test_basic.py
```diff
@@ -10,7 +10,7 @@
 import roswire.exceptions
 from dockerblade import Shell, FileSystem
 from roswire import AppInstance, ROSWire, AppDescription
-from roswire.proxy import ROS1
+from roswire.ros1 import ROS1
```
ChrisTimperley, line 13: We could add a `ROS1` and `ROS2` import to `roswire/__init__.py`?

schmerl:
Made all the changes suggested

src/roswire/ros1/__init__.py
ChrisTimperley: Looks like this line is still missing?

ChrisTimperley (approved):

PR #428, opened 2020-10-20
Moved contents of proxy/roslaunch into common submodule

schmerl:
This is in preparation for better refactoring between ROS1 and ROS2 parts.


src/roswire/proxy/ros1.py
```diff
@@ -3,20 +3,23 @@
...
 import xmlrpc.client
 from typing import Dict, Optional, Sequence, Tuple
 
 import dockerblade
 from loguru import logger
 
-from .roslaunch import ROSLaunchManager
-from ..app.description import AppDescription
 from ..common import NodeManager, SystemState
+from ..common.roslaunch import ROSLaunchManager
```
ChrisTimperley, line 14: It should be possible to merge this into the `from .. import` above.

ChrisTimperley (approved):
Just a minor change and then we're good to go 👍🏻

PR #430, opened 2020-10-20
Renamed "roslaunch" module to launch"

schmerl:
This is because roslaunch is a ROS1 command, which doesn't exist in ROS2 (it's `ros2 launch`)

Kept `ros1/launch/roslaunch.py` as roslaunch because it seems appropriate.


src/roswire/common/launch/controller.py
```diff
@@ -10,7 +10,7 @@
 
 @attr.s(frozen=True, slots=True, auto_attribs=True)
 class ROSLaunchController:
```
ChrisTimperley, line 12: Should we rename to `LaunchController`?

src/roswire/ros2/__init__.py
```diff
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
+from .launch.launch import ROS2LaunchManager
```
ChrisTimperley, line 2: We should be able to just to `from .launch import ROS2LaunchManager`.

src/roswire/ros2/ros2.py
```diff
@@ -5,8 +5,8 @@
 
 import attr
 
+from .launch.launch import ROS2LaunchManager
```
ChrisTimperley, line 8: Same as above

src/roswire/common/launch/controller.py
schmerl: I thought that could remain because it is the Launch Controller for all versions of ROS

src/roswire/ros2/launch/reader.py
```diff
@@ -43,7 +43,7 @@ def read(self,
             file inside the container.
         argv: Sequence[str], optional
             An optional sequence of command-line arguments that should be
-            supplied to :code:`roslaunch`.
+            supplied to :code:`launch`.
```
ChrisTimperley, line 46: `ros2 launch`

src/roswire/common/__init__.py
```diff
@@ -2,11 +2,11 @@
 from .action import ActionFormat
 from .base import Duration, Time
 from .format import FormatDatabase
+from .launch import ROSLaunchManager
```
ChrisTimperley, line 5: Should we rename to `LaunchManager`?

src/roswire/common/launch/controller.py
ChrisTimperley: That's fair.

src/roswire/common/__init__.py
schmerl: See above. I think this is ok.

schmerl: Otherwise, we'd have to name all the other ROSxxx things to be consistent, e.g., ROSNodeManager becomes NodeManager.

ChrisTimperley (approved):

PR #433, opened 2020-10-28
Refactor PackageDatabase building into a build method

schmerl:


src/roswire/common/package.py
```diff
@@ -193,10 +204,17 @@ def from_paths(app_instance: 'AppInstance',
                     raise
             else:
                 packages.append(package)
-        return PackageDatabase(packages)
+        return PackageDatabase(packages, paths)
 
-    def __init__(self, packages: Collection[Package]) -> None:
+    def __init__(self,
+                 packages: Collection[Package],
```
ChrisTimperley, line 210: Small optimisation: `packages` could actually be an `Iterable[Package]`, that would let us pass a generator instead (avoiding the need to create redundant lists).

src/roswire/common/package.py
```diff
@@ -193,10 +204,17 @@ def from_paths(app_instance: 'AppInstance',
...
+                 paths: List[str]) -> None:
```
ChrisTimperley, line 211: Same applies for `paths`.

src/roswire/common/package.py
```diff
@@ -221,7 +239,7 @@ def __iter__(self) -> Iterator[str]:
 
     @staticmethod
     def from_dict(d: List[Dict[str, Any]]) -> 'PackageDatabase':
-        return PackageDatabase([Package.from_dict(dd) for dd in d])
+        return PackageDatabase([Package.from_dict(dd) for dd in d], [])
```
ChrisTimperley, line 242: If you go with the optimisation suggested above, you could drop the list comprehension (for the first argument) and do `( ... )` instead to use a generator rather than a list.

test/test_package.py
```diff
@@ -237,7 +237,8 @@ def test_database_paths(sut):
         '/opt/ros/melodic/share/xacro',
         '/opt/ros/melodic/share/xmlrpcpp'
     }
-    actual = set(PackageDatabase.paths(sut))
+    pdb = PackageDatabase.build(sut)
```
ChrisTimperley, line 240: This probably ought to be `PackageDatabase._determine_paths`.

test/test_package.py
```diff
@@ -431,7 +432,7 @@ def test_package_location_ros2(sut):
                       '/ros_ws/install/launch',
                       '/ros_ws/install/ament_cmake_export_link_flags'
                       }
-    actual_paths = set(PackageDatabase.paths(sut))
+    actual_paths = set(PackageDatabase._determine_paths(sut))
```
ChrisTimperley, line 435: Whereas this one should be `db.paths` 🙂

test/test_package.py
```diff
@@ -431,7 +431,7 @@ def test_package_location_ros2(sut):
                       '/ros_ws/install/launch',
                       '/ros_ws/install/ament_cmake_export_link_flags'
                       }
-    actual_paths = set(PackageDatabase.paths(sut))
```
ChrisTimperley, line 434: I think this one is wrong. This should be `actual_paths = set(db.paths)`.

ChrisTimperley (approved):
👍🏻 LGTM

PR #438, opened 2020-11-12
Implemented param delete (fixes #387)

### LINKED ISSUE #387: Implement rosparam delete
I need to have `rosparam delete` implemented, because the Fetch demo on TheRobotCooperative has a launch file that uses it. Currently, roswire throws an exception, meaning all downstream processing is stopped/prevented.

Locally, I have a workaround that ignores `delete` if the parameter doesn't have a value anyway. So I have replaced [reader.py](https://github.com/ChrisTimperley/roswire/blob/00ec6f40b7c05babdf95fa6e734d1a56bf9cb73d/src/roswire/proxy/roslaunch/reader.py#L259-L261) with:

```python
        # handle delete command
        if cmd == 'delete':
            # Only raise an error if the param has a value
            if cfg.params.get(full_param) is not None:
                logger.error(f"Trying to delete {full_param}, whose value is {cfg.params.get(full_param)}")
                m = f"'delete' command is currently not supported in <rosparam> in {ctx.filename}"
                raise NotImplementedError(m)
```

But this is a temporary fix at best.

schmerl:
The fetch example uses param delete, so it is needed for rosdiscover.

### DIFF
--- a/src/roswire/common/launch/config/launch.py
+++ b/src/roswire/common/launch/config/launch.py
@@ -99,6 +99,20 @@ class LaunchConfig:
         params[name] = param
         return attr.evolve(self, params=params, errors=errors)
 
+    def without_param(self, name: str) -> "LaunchConfig":
+        params: Dict[str, Any] = dict(self.params)
+        errors = self.errors
+
+        if not name_is_global(name):
+            m = f"expected parameter name to be global: {name}"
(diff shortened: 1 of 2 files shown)


ChrisTimperley (approved):
This looks brilliant. Thanks! 👍🏻

PR #439, opened 2020-11-13
Added ability to attach to existing containers and improved ROS1 interface (fixes #1)

### LINKED ISSUE #1: ROSProxy should block until Master is reachable or timeout occurs

ChrisTimperley:


src/roswire/app/app.py
```diff
@@ -155,3 +155,39 @@ def launch(
...
+            If :code:`True`, this method call will ensure that there is a
+            description of the application before launching. If :code:`False`,
+            no check will be performed.
+
+        Returns
+        -------
+        AppInstance
+            A ROSWire interface around the given container.
+        """
+        dockerblade = self._roswire._dockerblade
```
schmerl, line 182: Is is proper to access a protected member of a `_roswire` in this way?

schmerl (commented):
Just a small questions in here.

src/roswire/app/app.py
ChrisTimperley: I think this is fine in this particular case. We don't necessarily want to expose `_dockerblade` to external clients of ROSWire, but internally that information is useful. This roughly fits the intuition behind Java's `protected` modifier.

PR #440, opened 2020-12-14
Prepared for separation of SystemState into ROS-specific versions.

schmerl:
The system state in ROS2 is more complete that ROS1, having actions and service clients recoverable as part of the node info. This pull request sets the refactoring of the SystemState into a version for ROS1 and ROS2.

### DIFF
--- a/src/roswire/ros1/state.py
+++ b/src/roswire/ros1/state.py
@@ -2,12 +2,52 @@
 __all__ = ("SystemStateProbe",)
 
 import xmlrpc
-from typing import Dict, Sequence, Tuple
+from typing import AbstractSet, Collection, Dict, Mapping, Sequence, Set, Tuple
 
 import attr
 
 from .. import exceptions as exc
 from ..common import SystemState
+
+
+@attr.s(frozen=True, auto_attribs=True, slots=True)
+class ROS1SystemState(SystemState):
+    """Provides a description of the instantaneous state of a ROS system in
+    terms of its publishers, subscribers, and services.
+
+    Attributes
+    ----------
+    publishers: Mapping[str, Collection[str]]
+        A mapping from topics to the names of publishers to that topic.
+    subscribers: Mapping[str, Collection[str]]
+        A mapping from topics to the names of subscribers to that topic.
+    services: Mapping[str, Collection[str]]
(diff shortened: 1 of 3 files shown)


src/roswire/ros2/state.py
```diff
@@ -14,6 +14,47 @@
     from .. import AppInstance
 
 
+@attr.s(frozen=True, auto_attribs=True, slots=True)
+class ROS2SystemState(SystemState):
```
schmerl, line 18: For the moment, this is the same as ROS1SystemState, which was the same as the old SystemState. Once this refactoring is pushed, I will start adding the ROS2 specific information here.

src/roswire/ros1/state.py
```diff
@@ -21,7 +61,7 @@ class SystemStateProbe:
 
     @classmethod
     def via_xmlrpc_connection(
-        cls, connection: xmlrpc.client.ServerProxy
+            cls, connection: xmlrpc.client.ServerProxy
```
ChrisTimperley, line 64: Minor nitpick: There's a little bit of excess indentation here, and it would be easier to read if `connection` were on another line.

src/roswire/ros2/state.py
ChrisTimperley: Would it make more sense to `raise UnimplementedError`?

src/roswire/ros1/state.py
```diff
@@ -2,14 +2,54 @@
...
 
 import attr
 
 from .. import exceptions as exc
 from ..common import SystemState
 
 
+@attr.s(frozen=True, auto_attribs=True, slots=True)
+class ROS1SystemState(SystemState):
+    """Provides a description of the instantaneous state of a ROS system in
```
ChrisTimperley, line 15: Let's lift the docstring into the `SystemState` interface and omit it from the implementing classes.

src/roswire/ros2/state.py
ChrisTimperley: Sorry, I was confused. The implementation is the same as ROS1 _for now_, but is duplicated here as it is going to change in a subsequent PR.

ChrisTimperley (approved):
LGTM 👍🏻

PR #442, opened 2020-12-17
Add clients and action servers/clients to ROS2 system state.

schmerl:


src/roswire/ros2/state.py
```diff
@@ -28,31 +28,61 @@ class ROS2SystemState(SystemState):
...
+    action_servers: Mapping[str, Collection[str]]
+        A mapping from actions to the names of providesr of that action.
+    action_clients: Mapping[str, Collection[str]]
+        A mapping from actions to the names of clients of that action
     nodes: AbstractSet[str]
         The names of all known nodes running on the system.
     topics: AbstractSet[str]
         The names of all known topics on the system with at least one
         publisher or one subscriber.
+    all_services: AbstractSet[str]
```
schmerl, line 42: This is a cumbersome and inconsistent name, but I can't call it `services` because that breaks the contract for SystemState.

ChrisTimperley: Can you call it `service_clients` to be consistent with `action_clients`?

src/roswire/ros2/state.py
```diff
@@ -28,31 +28,61 @@ class ROS2SystemState(SystemState):
         A mapping from topics to the names of subscribers to that topic.
     services: Mapping[str, Collection[str]]
         A mapping from services to the names of providers of that service.
+    clients: Mapping[str, Collection[str]]
```
ChrisTimperley, line 31: Can you call it `service_clients` to be consistent with `action_clients`?

src/roswire/ros2/state.py
ChrisTimperley: Maybe `service_names` will do for now?

src/roswire/ros2/state.py
```diff
@@ -28,31 +28,61 @@ class ROS2SystemState(SystemState):
...
+        A mapping from actions to the names of clients of that action
     nodes: AbstractSet[str]
         The names of all known nodes running on the system.
     topics: AbstractSet[str]
         The names of all known topics on the system with at least one
         publisher or one subscriber.
+    all_services: AbstractSet[str]
+        The name of all the known services on the system with at least
+        one server or client
+    actions: AbstractSet[str]
```
ChrisTimperley, line 45: Then this would become `action_names`, I suppose?

src/roswire/ros2/state.py
```diff
@@ -107,7 +140,14 @@ def probe(self) -> ROS2SystemState:
                     mode = "serv"
                     continue
                 elif "Action Servers:" in line:
-                    break
+                    mode = "act_serv"
```
ChrisTimperley, line 143: Should this be `act_serv` or `a_serv`?

src/roswire/ros2/state.py
```diff
@@ -107,7 +140,14 @@ def probe(self) -> ROS2SystemState:
...
+                    continue
+                elif "Service Clients:" in line:
+                    mode = "cli"
+                    continue
+                elif "Action Clients:" in line:
+                    mode = "act_cli"
```
ChrisTimperley, line 149: String bug?

src/roswire/ros2/state.py
```diff
@@ -75,6 +105,9 @@ def probe(self) -> ROS2SystemState:
             "pub": {},
             "sub": {},
             "serv": {},
+            "cli": {},
```
ChrisTimperley, line 108: It might help readability (and avoid some string bugs, see below) by changing these shorthand names to longer form equivalents (e.g., `service_clients`, `publishers`, `subscribers`, `service_providers`, etc.).

ChrisTimperley (changes requested):
We should improve a few API names and fix a likely string bug.

src/roswire/ros2/state.py
schmerl: Done

src/roswire/ros2/state.py
schmerl: Done

src/roswire/ros2/state.py
schmerl: Done

src/roswire/ros2/state.py
schmerl: Pulled these out as constants to avoid this

src/roswire/ros2/state.py
schmerl: See above

src/roswire/ros2/state.py
schmerl: See above

src/roswire/ros2/state.py
```diff
@@ -10,6 +10,13 @@
 
 from ..common import SystemState
 
+ACTION_CLIENTS = "act_cli"
```
ChrisTimperley, line 13: Minor nitpick: It might be worth adding a leading underscore to these variables to avoid them being included in the module's public API (i.e., `_ACTION_CLIENTS`).

src/roswire/ros2/state.py
ChrisTimperley: It looks like the docstring is still using the old name here

src/roswire/ros2/state.py
```diff
@@ -28,31 +35,61 @@ class ROS2SystemState(SystemState):
...
+        nodes = nodes.union(*self.service_clients.values())
+        nodes = nodes.union(*self.action_servers.values())
+        nodes = nodes.union(*self.action_clients.values())
 
         topics: Set[str] = set()
         topics = topics.union(self.publishers)
         topics = topics.union(self.subscribers)
 
+        service_names: Set[str] = set()
+        service_names = service_names.union(self.services)
```
ChrisTimperley, line 82: This might produce unexpected results. `self.services` has type `Mapping[str, Collection[str]]`, but `union` is expecting `Iterable[str]`.

src/roswire/ros2/state.py
```diff
@@ -28,31 +35,61 @@ class ROS2SystemState(SystemState):
...
+        service_names = service_names.union(self.service_clients)
+
+        action_names: Set[str] = set()
+        action_names = action_names.union(self.action_servers)
```
ChrisTimperley, line 86: Same as above.

src/roswire/ros2/state.py
ChrisTimperley: Nice

ChrisTimperley (approved):
This looks great! Let's get it merged 👍🏻

PR #443, opened 2021-01-06
Add topic, service, and action type information to ROS2State

schmerl:
The type information in ROS2 can be gleaned from ros2 info, and so we will cache it in the state.

### DIFF
--- a/src/roswire/exceptions.py
+++ b/src/roswire/exceptions.py
@@ -159,6 +159,18 @@ class ParsingError(ROSWireException):
 
 
 @_attr.s(frozen=True, auto_exc=True, auto_attribs=True, str=False)
+class ConflictingTypes(ROSWireException):
+    """ROSWire found two or more types for a service, topic, or action """
+
+    entity: str
+    existing: str
+    conflicting: str
+
+    def __str__(self) -> str:
+        return f"error declaring type {self.conflicting} for {self.entity}. It already has " \
+               f"type {self.existing}"
+
+@_attr.s(frozen=True, auto_exc=True, auto_attribs=True, str=False)
 class NoDescriptionError(RuntimeError, ROSWireException):
     """No description has been generated for an application."""
 
--- a/src/roswire/exceptions.py
+++ b/src/roswire/exceptions.py
@@ -167,8 +167,8 @@ class ConflictingTypes(ROSWireException):
     conflicting: str
 
     def __str__(self) -> str:
-        return f"error declaring type {self.conflicting} for {self.entity}. " \
-               f"It already has type {self.existing}"
+        return (f"error declaring type {self.conflicting} for {self.entity}. "
+                f"It already has type {self.existing}")
 
 
 @_attr.s(frozen=True, auto_exc=True, auto_attribs=True, str=False)
--- a/src/roswire/exceptions.py
+++ b/src/roswire/exceptions.py
@@ -160,15 +160,16 @@ class ParsingError(ROSWireException):
 
 @_attr.s(frozen=True, auto_exc=True, auto_attribs=True, str=False)
 class ConflictingTypes(ROSWireException):
-    """ROSWire found two or more types for a service, topic, or action """
+    """ROSWire found two or more types for a service, topic, or action."""
 
     entity: str
     existing: str
     conflicting: str
 
     def __str__(self) -> str:
-        return f"error declaring type {self.conflicting} for {self.entity}. It already has " \
-               f"type {self.existing}"
+        return f"error declaring type {self.conflicting} for {self.entity}. " \
+               f"It already has type {self.existing}"
+
 
 @_attr.s(frozen=True, auto_exc=True, auto_attribs=True, str=False)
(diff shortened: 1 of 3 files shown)


src/roswire/ros2/state.py
```diff
@@ -41,6 +41,12 @@ class ROS2SystemState(SystemState):
         A mapping from actions to the names of providesr of that action.
     action_clients: Mapping[str, Collection[str]]
         A mapping from actions to the names of clients of that action
+    topic_to_type: Mapping[str, str]
+        A mapping from topics to their type
+    service_to_type: Mapping[str, str]
+        A mapping from services to their type
+    action_to_type:
```
ChrisTimperley, line 48: Missing type annotation

src/roswire/ros2/state.py
```diff
@@ -41,6 +41,12 @@ class ROS2SystemState(SystemState):
         A mapping from actions to the names of providesr of that action.
     action_clients: Mapping[str, Collection[str]]
         A mapping from actions to the names of clients of that action
+    topic_to_type: Mapping[str, str]
+        A mapping from topics to their type
```
ChrisTimperley, line 45: For clarity, we should change `to their type` to `to the name of their type`

src/roswire/ros2/state.py
```diff
@@ -116,6 +125,11 @@ def probe(self) -> ROS2SystemState:
             _ACTION_SERVERS: {},
             _ACTION_CLIENTS: {}
         }
+        # The place to store type information
```
ChrisTimperley, line 128: I don't think that the comment is really necessary here

src/roswire/ros2/state.py
```diff
@@ -139,37 +154,57 @@ def probe(self) -> ROS2SystemState:
...
                     if name in node_to_state[mode]:
                         node_to_state[mode][name].add(node_name)
                     else:
                         node_to_state[mode][name] = {node_name}
+                    # Add type information and report conflict if a
+                    # different types was registered (probably
+                    # should never happen)
+                    if name in types and fmt != types[name]:
+                        logger.warning(
+                            f'The entity {name} has conflictig types: '
```
ChrisTimperley, line 193: typo

src/roswire/ros2/state.py
```diff
@@ -139,37 +154,57 @@ def probe(self) -> ROS2SystemState:
...
+                    name = partition[0]
+                    fmt = partition[1]
                     if name in node_to_state[mode]:
                         node_to_state[mode][name].add(node_name)
                     else:
                         node_to_state[mode][name] = {node_name}
+                    # Add type information and report conflict if a
+                    # different types was registered (probably
+                    # should never happen)
+                    if name in types and fmt != types[name]:
```
ChrisTimperley, line 191: This is a nice sanity check, but I think we should either raise an exception or use an assertion here. (Continuing with the execution could lead to strange results.)

schmerl: Chose to raise an exception.

src/roswire/ros2/state.py
schmerl: Done

src/roswire/ros2/state.py
schmerl: Done

src/roswire/ros2/state.py
schmerl: OK.

src/roswire/ros2/state.py
schmerl: Done,

src/roswire/ros2/state.py
schmerl: Raising an exception

src/roswire/exceptions.py
```diff
@@ -158,6 +158,19 @@ class ParsingError(ROSWireException):
...
+@_attr.s(frozen=True, auto_exc=True, auto_attribs=True, str=False)
+class ConflictingTypes(ROSWireException):
+    """ROSWire found two or more types for a service, topic, or action."""
+
+    entity: str
+    existing: str
+    conflicting: str
+
+    def __str__(self) -> str:
+        return f"error declaring type {self.conflicting} for {self.entity}. " \
```
ChrisTimperley, line 170: Prefer the use of parens to break up strings over `\`.
E.g.,
```
("foo"
 "bar")
```

src/roswire/ros2/state.py
```diff
@@ -106,7 +116,13 @@ def for_app_instance(cls, app_instance: "AppInstance") -> "ROS2StateProbe":
         return ROS2StateProbe(app_instance=app_instance)
 
     def probe(self) -> ROS2SystemState:
-        """Obtains the instantaneous state of the associated ROS system."""
+        """Obtains the instantaneous state of the associated ROS system.
+
+        Raises
+        ------
+        ConflictingTypeException
+            If more than one type is detected for topics, services, or actions
```
ChrisTimperley, line 124: slight tweak: `is detected for a given topic, service, or action.`

ChrisTimperley (approved):
👍🏻 LGTM

PR #444, opened 2021-01-11
Integrate ROS2State with ROSWire system

schmerl:
Integrate in the use of ROS2 System State so that clients can access the ROS2 information.

### DIFF
--- a/src/roswire/ros2/package.py
+++ b/src/roswire/ros2/package.py
@@ -3,6 +3,7 @@ __all__ = ("ROS2PackageDatabase",)
 
 import json
 import os
+import re
 import typing
 from typing import Any, Dict, Iterable, List, Mapping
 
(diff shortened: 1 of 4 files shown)


src/roswire/ros2/package.py
```diff
@@ -23,6 +24,9 @@
     "'"
 )
 
+_COMMAND_ROS2_PKG_PREFIXES_3: Final[str] = re.compile("python ").\
```
ChrisTimperley, line 27: This seems like a work in progress?

src/roswire/ros2/package.py
```diff
@@ -42,8 +46,13 @@ def _determine_paths(cls, app_instance: "AppInstance") -> List[str]:
             shell = app_instance.shell
             jsn = shell.check_output(_COMMAND_ROS2_PKG_PREFIXES, text=True)
         except dockerblade.exceptions.CalledProcessError:
-            logger.error("failed to obtain ROS2 package prefixes")
-            raise
+            # maybe python is python3, try that
```
ChrisTimperley, line 49: Ah. I think I see what's happening here. I think we can just use `python3` rather than `python` when we're working with ROS2.

schmerl: Done

ChrisTimperley (approved):
👍🏻

PR #446, opened 2021-01-19
Fixed minor bugs in launch controller and roslaunch

afsafzal:


ChrisTimperley (approved):
Thanks for fixing these! 👍🏻

Issue #420, opened 2020-10-16
Fields in message files may provide optional default values


ChrisTimperley:
The current message file parser doesn't recognise this, and so it crashes on any message fields that include optional default values.

https://design.ros2.org/articles/legacy_interface_definition.html

schmerl:
To get a start on this, I think we need to:

1. Refactor `common/package.py` to make Package an interface
2. Move the current implementation into ROS1Package
3. Hook this in
4. Split `MsgFromat`, `ActionFormat`, and `SrvFormat` into interface and ROS1 / ROS2 instances.
5. Hook them into ROS1Package
6. Create ROS2Package that uses the new formats
7. Hook this into the ROS2 specific stuff

schmerl:
Look at: https://github.com/ros2/rosidl/blob/master/rosidl_adapter/rosidl_adapter/parser.py

schmerl:
https://index.ros.org/doc/ros2/Concepts/About-ROS-Interfaces/

schmerl:
#394

PR #447, opened 2021-01-22
Refactor Package into ROS1 and ROS2 versions

schmerl:
(Note currently ROS2Package and ROS1Package are identical and will be changed in subsequent tasks of #420)

### DIFF
--- a/src/roswire/common/package.py
+++ b/src/roswire/common/package.py
@@ -12,7 +12,6 @@ from typing import (
     List,
     Mapping,
     Optional,
-    Sequence
 )
 
 import attr
@@ -81,7 +80,9 @@ class PackageDatabase(Generic[PT], ABC, Mapping[str, PT]):
     _contents: Mapping[str, PT] = attr.ib()
 
     @classmethod
-    def from_packages(cls, packages: typing.Iterable[PT]) -> "PackageDatabase[PT]":
+    def from_packages(cls,
+                      packages: typing.Iterable[PT]
+                      ) -> "PackageDatabase[PT]":
         return cls({p.name: p for p in packages})
 
     @classmethod
--- a/src/roswire/common/package.py
+++ b/src/roswire/common/package.py
@@ -1,93 +1,46 @@
 # -*- coding: utf-8 -*-
 __all__ = ("Package", "PackageDatabase")
 
-import os
 import typing
 from abc import ABC, abstractmethod
 from typing import (
     Any,
+    Collection,
     Dict,
-    Iterable,
-    Iterator,
+    Generic,
     List,
     Mapping,
-    Optional,
-    Tuple,
+    Optional
 )
-
-import attr
-from loguru import logger
 
 from .action import ActionFormat
 from .msg import MsgFormat
 from .srv import SrvFormat
-from ..util import tuple_from_iterable
 
 if typing.TYPE_CHECKING:
     from .. import AppInstance
 
+MF = typing.TypeVar("MF", bound=MsgFormat)
+SF = typing.TypeVar("SF", bound=SrvFormat)
+AF = typing.TypeVar("AF", bound=ActionFormat)
 
-@attr.s(frozen=True, auto_attribs=True, slots=True)
-class Package:
+
+class Package(Generic[MF, SF, AF], ABC):
     name: str
     path: str
-    messages: Tuple[MsgFormat, ...] = attr.ib(converter=tuple_from_iterable)
-    services: Tuple[SrvFormat, ...] = attr.ib(converter=tuple_from_iterable)
-    actions: Tuple[ActionFormat, ...] = attr.ib(converter=tuple_from_iterable)
+    messages: Collection[MF]
+    services: Collection[SF]
+    actions: Collection[AF]
 
-    @staticmethod
-    def build(path: str, app_instance: "AppInstance") -> "Package":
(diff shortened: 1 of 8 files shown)


src/roswire/common/package.py
```diff
@@ -116,116 +69,19 @@ class PackageDatabase(ABC, Mapping[str, Package]):
...
 
-    @classmethod
-    def build(
-        cls, app_instance: "AppInstance", paths: Optional[List[str]] = None
-    ) -> "PackageDatabase":
-        if paths is None:
-            paths = cls._determine_paths(app_instance)
-        db_package = cls._from_paths(app_instance, paths)
-        return db_package
+    _paths_in_package: List[str]
```
ChrisTimperley, line 72: If this is intended as a protected attribute, it should move to the same location as the other attributes. Also, does the type annotation need to be `List`? Or is `Sequence` or `Collection` sufficient?

src/roswire/common/package.py
```diff
@@ -116,116 +69,19 @@ class PackageDatabase(ABC, Mapping[str, Package]):
...
-
-        Parameters
-        ----------
-        app_instance: AppInstance
-            An instance of the application for which the
-            list of paths should be obtained
-        """
+    def build(cls,
+              app_instance: "AppInstance",
+              paths: Optional[List[str]]) -> "PackageDatabase":
```
ChrisTimperley, line 78: Nitpick: return type should be on a separate line.
More importantly, `paths` should have a default value of `None`.

src/roswire/common/package.py
```diff
@@ -116,116 +69,19 @@ class PackageDatabase(ABC, Mapping[str, Package]):
...
-                packages.append(package)
-        return cls._from_packages_and_paths(packages, paths)
-
-    def __init__(
-        self, packages: Iterable[Package], paths: Iterable[str]
-    ) -> None:
-        self.__contents = {p.name: p for p in packages}
-        self._paths_in_package = list(paths)
-
     @property
```
ChrisTimperley, line 81: If we have `paths` as an immutable attribute, we won't need this getter.

src/roswire/ros1/package.py
```diff
@@ -3,27 +3,160 @@
...
-        cls, packages: Iterable[Package], paths: Iterable[str]
-    ) -> "PackageDatabase":
+            cls, packages: Iterable[ROS1Package], paths: Iterable[str]
+    ) -> "ROS1PackageDatabase":
         return ROS1PackageDatabase(packages, paths)
 
     @classmethod
-    def from_dict(cls, d: List[Dict[str, Any]]) -> "PackageDatabase":
-        return ROS1PackageDatabase([Package.from_dict(dd) for dd in d], [])
+    def from_dict(cls, d: List[Dict[str, Any]]) -> "ROS1PackageDatabase":
```
ChrisTimperley, line 104: Return type should be `ROSPackageDatabase` to avoid breaking LSP.

src/roswire/ros1/package.py
```diff
@@ -3,27 +3,160 @@
...
+            "messages": [m.to_dict() for m in self.messages],
+            "services": [s.to_dict() for s in self.services],
+            "actions": [a.to_dict() for a in self.actions],
+        }
+        return d
+
+
 class ROS1PackageDatabase(PackageDatabase):
+
+    __contents: Dict[str, Any]
```
ChrisTimperley, line 95: I'm not entirely sure what this is.

src/roswire/common/package.py
schmerl: The property paths() accesses this, so I don't think I can move it. (Or I need to duplicate it down in all subclasses.)

src/roswire/ros2/package.py
```diff
@@ -50,3 +179,30 @@ def _determine_paths(cls, app_instance: "AppInstance") -> List[str]:
             for (package, prefix) in package_to_prefix.items()
         ]
         return paths
+
+    def __init__(
+            self, packages: Iterable[ROS2Package], paths: Iterable[str]
```
ChrisTimperley, line 184: Excessive indentation. Readability could be improved by using a separate line for each argument.

src/roswire/common/package.py
schmerl: Done

src/roswire/ros1/package.py
```diff
@@ -1,29 +1,110 @@
...
+@attr.s(frozen=True, auto_attribs=True, slots=True)
+class ROS1Package(Package[MsgFormat, SrvFormat, ActionFormat]):
+    name: str
+    path: str
+    messages: Collection[MsgFormat] = attr.ib(converter=tuple_from_iterable)
+    services: Collection[SrvFormat] = attr.ib(converter=tuple_from_iterable)
+    actions: Collection[ActionFormat] = attr.ib(converter=tuple_from_iterable)
+
+    @classmethod
+    def build(cls, path: str, app_instance: "AppInstance") -> "ROS1Package":
```
ChrisTimperley, line 30: This violates LSP. (Return type of `build` in `Package` is `Package`.) We'll need to think about how to deal with this one.

ChrisTimperley: One way to avoid this issue is to move this logic into the `_build_package` method of `PackageDatabase`.

ChrisTimperley (approved):
👍🏻 Awesome. Let's get it merged!

PR #448, opened 2021-01-26
Refactor pattern matching in Msgs into constituent parts

schmerl:
Move the regular expression matching into Constant and Field. This will help down the line with constructing ROS2Field with different pattern for default and bounded.

(This is related to issue #420)

### DIFF
--- a/src/roswire/common/msg.py
+++ b/src/roswire/common/msg.py
@@ -29,13 +29,8 @@ from .base import Duration, is_builtin, Time
 from .decode import is_simple
 from .. import exceptions as exc
 
-R_TYPE = r"[a-zA-Z0-9_/]+(?:\[(?:<=)?\d*\])?"
-R_NAME = r"[a-zA-Z0-9_/]+"
-R_VAL = r".+"
+
 R_COMMENT = r"(#.*)?"
-R_FIELD = re.compile(f"^\s*({R_TYPE})\s+({R_NAME})\s*{R_COMMENT}$")
-R_STRING_CONSTANT = re.compile("^\s*string\s+(\w+)\s*=\s*(.+)\s*$")
-R_OTHER_CONSTANT = re.compile("^\s*(\w+)\s+(\w+)\s*=\s*([^\s]+).*$")
 R_BLANK = re.compile(f"^\s*{R_COMMENT}$")
 
 ConstantValue = Union[str, int, float]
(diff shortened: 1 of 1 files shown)


src/roswire/common/msg.py
```diff
@@ -55,10 +50,42 @@ class Constant:
...
 
+    R_STRING_CONSTANT = re.compile("^\s*string\s+(\w+)\s*=\s*(.+)\s*$")
+    R_OTHER_CONSTANT = re.compile("^\s*(\w+)\s+(\w+)\s*=\s*([^\s]+).*$")
+
     typ: str
     name: str
     value: Union[str, int, float]
 
+    @classmethod
+    def from_string(cls, package: str, line: str) -> "Optional[Constant]":
```
ChrisTimperley, line 61: It looks like `package` isn't used here?

src/roswire/common/msg.py
```diff
@@ -82,9 +109,51 @@ class Field:
         The name of this field.
     """
 
+    R_TYPE = r"[a-zA-Z0-9_/]+(?:\[(?:<=)?\d*\])?"
+    R_NAME = r"[a-zA-Z0-9_/]+"
+    R_VAL = r".+"
```
ChrisTimperley, line 114: Is `R_VAL` used here?

src/roswire/common/msg.py
```diff
@@ -29,13 +29,8 @@
 from .decode import is_simple
 from .. import exceptions as exc
 
-R_TYPE = r"[a-zA-Z0-9_/]+(?:\[(?:<=)?\d*\])?"
-R_NAME = r"[a-zA-Z0-9_/]+"
-R_VAL = r".+"
+
 R_COMMENT = r"(#.*)?"
```
ChrisTimperley, line 33: Does it make sense to move the remaining regexes below?

src/roswire/common/msg.py
```diff
@@ -55,10 +50,42 @@ class Constant:
...
+    R_OTHER_CONSTANT = re.compile("^\s*(\w+)\s+(\w+)\s*=\s*([^\s]+).*$")
+
     typ: str
     name: str
     value: Union[str, int, float]
 
+    @classmethod
+    def from_string(cls, package: str, line: str) -> "Optional[Constant]":
+        """
+
```
ChrisTimperley, line 63: Missing one-line method description here and below.

src/roswire/common/msg.py
```diff
@@ -55,10 +50,42 @@ class Constant:
...
+        Parameters
+        ----------
+        package: str
+            The name of the package that provides the constant.
+        line: str
```
ChrisTimperley, line 68: Bit of a nitpicky comment. Some of these lines end with a period whereas others don't. I have no particular preference, but we should probably be consistent.

src/roswire/common/msg.py
```diff
@@ -227,37 +296,14 @@ def from_string(package: str, name: str, text: str) -> "MsgFormat":
...
-            elif m_string_constant:
-                name_const, val = m_string_constant.group(1, 2)
-                constant = Constant("string", name_const, val)
-                constants.append(constant)
-            elif m_other_constant:
-                typ, name_const, val_str = m_other_constant.group(1, 2, 3)
-                val = val_str  # FIXME convert value
-                constant = Constant(typ, name_const, val)
+
+            constant = Constant.from_string(package, line)
```
ChrisTimperley, line 302: This is neat!

src/roswire/common/msg.py
schmerl: Right. I was doing it to make it consistent with the other from_string that is needed in Field. I also wasn't sure if constants can have types in the same way that Fields can, where you need to resolve a name in the package. (Currently, Constant has a todo on processing the type.)

src/roswire/common/msg.py
schmerl: Huh. Good point. I thought it was. It will be in ROS 2 but that's a different class.

src/roswire/common/msg.py
schmerl: `R_COMMENT` is used in both `Field` and and `MsgFormat`. I think leaving them as global for now is fine. And I'll keep them in the common package for the subclasses too.

src/roswire/common/msg.py
schmerl: Done.

src/roswire/common/msg.py
schmerl: Done.

src/roswire/common/msg.py
ChrisTimperley: I don't believe that the name of the package is necessary to determine the qualified name of the constant type. (I might be wrong for ROS2?) The FIXME rightly complains about the lack of type conversion in the constant value itself (e.g., integer types aren't converted from `str` to `int`), but I suppose that's a separate issue 🙂

schmerl: It looks like constants even in ROS 2 can only have primitive types. So I've removed the package parameter.

ChrisTimperley (approved):
This looks great. Let's get it merged! 👍🏻

PR #449, opened 2021-02-02
Refactor FormatDatabase into ROS1 and ROS2 versions

schmerl:
FormatDatabase depends on MsgFormat (and classes that depend on MsgFormat). To prepare for processing ROS2 message formats, this PR splits FormatDatabase first.

Note that ROS1FormatDatabase and ROS2FormatDatabase are mostly identical, but will change when ActionFormat and MsgFormat are refactored into different ROS versions.

### DIFF
--- a/src/roswire/common/format.py
+++ b/src/roswire/common/format.py
@@ -65,15 +65,6 @@ class FormatDatabase(ABC, Generic[MF, SF, AF]):
 
         return cls(messages, services, actions)
 
-    @classmethod
-    @abstractmethod
-    def build(cls,
-              messages: Set[MF],
-              services: Set[SF],
-              actions: Set[AF]
-              ) -> "FormatDatabase":
-        ...
-
     def __init__(
             self,
             messages: Set[MF],
@@ -116,6 +107,7 @@ class FormatDatabase(ABC, Generic[MF, SF, AF]):
             yaml.dump(self.to_dict(), f, default_flow_style=False)
 
     @classmethod
+    @abstractmethod
     def from_dict(cls, d: Dict[str, Any]) -> "FormatDatabase":
         """Loads a format database from a JSON document."""
         ...
--- a/src/roswire/common/format.py
+++ b/src/roswire/common/format.py
@@ -66,10 +66,10 @@ class FormatDatabase(ABC, Generic[MF, SF, AF]):
         return cls(messages, services, actions)
 
     def __init__(
-            self,
-            messages: Set[MF],
-            services: Set[SF],
-            actions: Set[AF],
+        self,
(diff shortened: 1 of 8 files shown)


src/roswire/ros1/format.py
```diff
@@ -0,0 +1,88 @@
+# -*- coding: utf-8 -*-
```
schmerl, line 1: Note that this file and ROS2FormatDatabase will be changed to use ROS-version specific MsgFromat and ActionFormat in a subsequent PR.

src/roswire/ros2/format.py
```diff
@@ -0,0 +1,88 @@
+# -*- coding: utf-8 -*-
```
schmerl, line 1: Note that this file and ROS2FormatDatabase will be changed to use ROS-version specific MsgFromat and ActionFormat in a subsequent PR.

src/roswire/common/format.py
```diff
@@ -30,85 +35,46 @@ class FormatDatabase:
         An immutable mapping from action name to definition.
     """
 
-    @staticmethod
-    def build(db: PackageDatabase) -> "FormatDatabase":
+    @classmethod
+    @abstractmethod
+    def build(cls, db: PackageDatabase) -> "FormatDatabase":
```
ChrisTimperley, line 40: Maybe change this to `from_packages(cls, db: PackageDatabase)` and add a `build(cls, messages, services, actions)` method.

src/roswire/common/format.py
```diff
@@ -56,57 +63,63 @@ def build(db: PackageDatabase) -> "FormatDatabase":
...
+              ) -> "FormatDatabase":
+        ...
 
     def __init__(
-        self,
-        messages: Set[MsgFormat],
-        services: Set[SrvFormat],
-        actions: Set[ActionFormat],
+            self,
+            messages: Set[MF],
```
ChrisTimperley, line 79: No need for the extra whitespace

src/roswire/common/format.py
```diff
@@ -56,57 +63,63 @@ def build(db: PackageDatabase) -> "FormatDatabase":
                 if action.feedback:
                     messages.add(action.feedback)
 
-        return FormatDatabase(messages, services, actions)
+        return cls(messages, services, actions)
```
ChrisTimperley, line 66: We probably want to either: (a) change this to `cls.build(...)`, or (b) remove `build` entirely since it's really just an alias for the constructor.

schmerl: Good point. I have removed the build method

ChrisTimperley (approved):
👍🏻 LGTM

PR #451, opened 2021-02-11
Separate ActionFormat into ROS1 and ROS2 versions

schmerl:
For the moment, functionality is duplicated across ROS1 and ROS2 implementation. Related to #420

### DIFF
--- a/src/roswire/common/action.py
+++ b/src/roswire/common/action.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 __all__ = ("ActionFormat",)
 
+import os
 from abc import ABC, abstractmethod
 from typing import Any, Dict, Generic, Optional, TypeVar
 
@@ -42,9 +43,8 @@ class ActionFormat(ABC, Generic[MF]):
     result: Optional[MF]
 
     @classmethod
-    @abstractmethod
     def from_file(
-        cls, package: str, filename: str, files: dockerblade.FileSystem
+            cls, package: str, filename: str, files: dockerblade.FileSystem
     ) -> "ActionFormat":
         """Constructs an action format from a .action file for a given package.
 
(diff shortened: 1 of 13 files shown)


src/roswire/common/action.py
```diff
@@ -25,26 +27,26 @@ class ActionFormat:
...
+    definition: str = attr.ib()
+    goal: MF = attr.ib()
+    feedback: Optional[MF] = attr.ib()
+    result: Optional[MF] = attr.ib()
 
-    @staticmethod
+    @classmethod
     def from_file(
-        package: str, filename: str, files: dockerblade.FileSystem
+            cls, package: str, filename: str, files: dockerblade.FileSystem
```
ChrisTimperley, line 49: Excess whitespace

src/roswire/common/action.py
```diff
@@ -79,52 +82,14 @@ def from_string(package: str, name: str, s: str) -> "ActionFormat":
...
-        return ActionFormat(package, name, s, goal, feed, res)
-
-    @staticmethod
+        ...
+
+    @classmethod
+    @abstractmethod
     def from_dict(
-        d: Dict[str, Any], *, package: Optional[str] = None
+            cls, d: Dict[str, Any], *, package: Optional[str] = None
```
ChrisTimperley, line 90: excess whitespace

test/test_format.py
```diff
@@ -9,9 +9,8 @@
     Field,
     MsgFormat,
     SrvFormat,
-    ActionFormat,
 )
-from roswire.ros1 import ROS1PackageDatabase, ROS1FormatDatabase
+from roswire.ros1 import ROS1ActionFormat, ROS1PackageDatabase, ROS1FormatDatabase
```
ChrisTimperley, line 13: prefer relative import

ChrisTimperley (changes requested):
A few minor stylistic points, but otherwise this looks good 👍🏻

test/test_format.py
schmerl: I think because this is a test case, we should do the `roswire.`, which is consistent with the other test cases.

PR #454, opened 2021-02-24
Separate SrvFormat into ROS1 and ROS2 versions #451

schmerl:
For the moment, the ROS1 and ROS2 functionality are the same. Related to #420

### DIFF
--- a/test/test_format.py
+++ b/test/test_format.py
@@ -7,9 +7,9 @@ import os
 from roswire.common import (
     Constant,
     Field,
-    MsgFormat,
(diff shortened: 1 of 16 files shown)


test/test_format.py
```diff
@@ -7,10 +7,10 @@
 from roswire.common import (
     Constant,
     Field,
-    MsgFormat,
-    SrvFormat,
+    MsgFormat,,
```
ChrisTimperley, line 10: typo?

PR #459, opened 2021-03-18
Miscellaneous fixes for using with ROSDiscover

schmerl:
The changes here are:

1. Support for bounded strings in messages (i.e., `string<=255`)
2. Support for ROS2 base type `wstring`
3. Fix in bug that didn't copy `launch_extractor._py` (needed for ROS2) into the distribution


src/roswire/ros2/msg.py
```diff
@@ -35,6 +38,16 @@ def from_string(cls, package: str, line: str) -> Optional["ROS2Field"]:
             return field
         return None
 
+    @classmethod
+    def _resolve_type(cls, package: str, typ: str) -> str:
+
+        # The string bounds (string<=123) will be removed to
+        # help resolution of the type
+        r_type = cls.REXP_TYPE.match(typ)
```
ChrisTimperley, line 46: prefer `match` to indicate that the result of the method call is an `re.Match` rather than a regular expression

src/roswire/ros2/msg.py
```diff
@@ -35,6 +38,16 @@ def from_string(cls, package: str, line: str) -> Optional["ROS2Field"]:
...
+        if r_type and r_type.group('strbounds'):
+            typ = typ.replace(r_type.group('strbounds'), '')
+        return Field._resolve_type(package, typ)
```
ChrisTimperley, line 49: Prefer `super()._resolve_type(package, typ)`

test/test_message.py
```diff
@@ -28,3 +29,12 @@ def test_to_and_from_dict(app: roswire.App):
...
         "position": {"x": position.x, "y": position.y, "z": position.z},
     }
+
+
+def test_from_string_ros2():
+    msg_string = ("string<=256 node_namespace\n"
+                  "string<=256 node_name\n"
+                  "Gid[] reader_gid_seq\n"
+                  "Gid[] writer_gid_seq\n")
+    msg = ROS2MsgFormat.from_string("", "", msg_string)
```
ChrisTimperley, line 39: Maybe worth checking the reported types of these fields?

ChrisTimperley (changes requested):
Just a few minor changes, but otherwise this looks good to merge! 👍🏻

PR #465, opened 2021-07-27
Switch from Travis to GitHub Actions

ChrisTimperley:


.github/workflows/mypy-flake8.yml
```diff
@@ -0,0 +1,27 @@
+name: "MyPy + Flake8"
+on: [push, pull_request]
```
schmerl, line 2: Just on PR?

schmerl (commented):
Just one small comment. otherwise LGTM!

schmerl:
Looks like need to install types for mypy?

PR #466, opened 2021-08-10
WIP: Integrate node source identification for packages

schmerl:


src/roswire/common/package.py
```diff
@@ -12,18 +15,85 @@
...
 
+
+class SourceLanguage(enum.Enum):
+    CXX = "cxx"
+    PYTHON = "python"
+
+
+@attr.s(frozen=True, auto_attribs=True, slots=True)
+class NodeSourceInfo:
+    node_name: Optional[str]
```
ChrisTimperley, line 42: There shouldn't be a case where a node doesn't have a name right?

src/roswire/ros1/package.py
```diff
@@ -46,6 +48,7 @@ def build(cls, path: str, app_instance: "AppInstance") -> "ROS1Package":
         dir_msg = os.path.join(path, "msg")
         dir_srv = os.path.join(path, "srv")
         dir_action = os.path.join(path, "action")
+        file_cmakelists = os.path.join(path, "CMakelists.txt")
```
ChrisTimperley, line 51: This should be `"CMakeLists.txt"`

schmerl: N/A

src/roswire/common/package.py
schmerl: Done.

schmerl:
Note. This image refactors the package extractor out of package and into its own class, and does implementation for ROS1 and partial for ROS2.

ROS2 uses both `CMakeLists.txt` and `setup.py` - we process the former and warn that we don't process the latter.

src/roswire/common/source.py
```diff
@@ -0,0 +1,139 @@
...
+    def from_dict(cls, info: t.Dict[str, t.Any]) -> "NodeSourceInfo":
+        return NodeSourceInfo(
+            info["name"],
+            SourceLanguage(info["language"]),
+            info["sources"]
+        )
+
+
+def extract_sources_from_cmake(
+        file_contents: str
```
ChrisTimperley, line 54: nitpick: indentation should be four spaces here.

src/roswire/common/source.py
```diff
@@ -0,0 +1,139 @@
...
+    @classmethod
+    def from_dict(cls, info: t.Dict[str, t.Any]) -> "NodeSourceInfo":
+        return NodeSourceInfo(
+            info["name"],
+            SourceLanguage(info["language"]),
+            info["sources"]
+        )
+
+
+def extract_sources_from_cmake(
```
ChrisTimperley, line 53: should this be a protected (helper) method in PackageSourceExtractor? That would help to hide this from the public API.

src/roswire/ros1/ros1.py
```diff
@@ -394,3 +405,28 @@ def playback(
             files=self.__files,
             delete_file_after_use=delete_file_after_use,
         )
+
+    def get_node_sources_for_source_package(
```
ChrisTimperley, line 409: name of this method is a bit lengthy. how about `node_sources_for_package` or `package_node_sources`?

src/roswire/ros2/ros2.py
```diff
@@ -23,16 +25,24 @@ class ROS2:
     services: ROS2ServiceManager = attr.ib(init=False)
     launch: ROS2LaunchManager = attr.ib(init=False)
     _state_probe: ROS2StateProbe = attr.ib(init=False)
+    __package_source_extractor: ROS2PackageSourceExtractor = \
```
ChrisTimperley, line 28: If I remember correctly, `__foo` causes issues with `attr.ib`? It might be worth just sticking to a single underscore here if this causes problems.

ChrisTimperley (approved):
Just a few minor comments about readability and one about class/method structure. Other than those, this looks good to merge 👍🏻

src/roswire/common/source.py
schmerl: Yeah. Its something I always have to remember to undo when I use autoformatting because I can't work out how to get it to format to only 4 spaces in argument lists.

src/roswire/common/source.py
schmerl: I don't think so. then there would only be a method to build it, and no method that does anything...?

src/roswire/ros2/ros2.py
schmerl: Done

src/roswire/ros2/source.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def extract_source_for_package(
+        self,
+        path_to_package: str
+    ) -> t.Mapping[str, NodeSourceInfo]:
+        cmakelists_path = os.path.join(path_to_package, "CMakeLists.txt")
+
+        if self._files.isfile(cmakelists_path):
+            contents = self._files.read(cmakelists_path)
+            source_infos = extract_sources_from_cmake(contents)
```
ChrisTimperley, line 38: e.g., you would call `self._extract_sources_from_cmake` here. `_extract_sources_from_cmake` can be defined as a static or class method in `PackageSourceExtractor`.

src/roswire/ros1/ros1.py
schmerl: Done.

src/roswire/common/source.py
schmerl: Done

src/roswire/common/source.py
ChrisTimperley: After discussion, totally agree with you here. We need this to be a public method for cases where we want to parse a CMakeLists.txt but we don't necessarily have a container (e.g., for rosanatomy).

PR #470, opened 2021-09-07
Allow recursive processing of CMakeLists.txt to include subdirectories

schmerl:


src/roswire/common/source.py
```diff
@@ -29,100 +32,70 @@ class SourceLanguage(enum.Enum):
     PYTHON = "python"
 
 
-@attr.s(frozen=True, auto_attribs=True, slots=True)
-class NodeSourceInfo:
-    node_name: str
+class ExecutableKind(enum.Enum):
+    NODE = "node"
+    LIBRARY = "library"
```
ChrisTimperley, line 37: A library isn't really a kind of executable, but both binaries and libraries are considered _build targets_.

src/roswire/common/source.py
```diff
@@ -29,100 +32,70 @@ class SourceLanguage(enum.Enum):
...
+
+
+@attr.s(auto_attribs=True, slots=True)
+class ExecutableInfo:
+    name: t.Optional[str]
```
ChrisTimperley, line 42: We probably need to think about the difference between the target name and the target's relative install destination.

src/roswire/common/source.py
```diff
@@ -29,100 +32,70 @@ class SourceLanguage(enum.Enum):
...
+    def from_dict(cls, info: t.Dict[str, t.Any]) -> "ExecutableInfo":
+        return ExecutableInfo(info["name"],
+                              SourceLanguage(info["language"]),
+                              ExecutableKind(info["kind"]),
+                              set(info["sources"]),
+                              set(info["path_restrictions"]))
 
+    @property
+    def entrypoint(self) -> t.Optional[str]:
+        if self.language == SourceLanguage.CXX:
```
ChrisTimperley, line 65: We'll want to specify entrypoints for libraries (e.g., nodelets)

src/roswire/common/source.py
```diff
@@ -29,100 +32,70 @@ class SourceLanguage(enum.Enum):
...
-                                SourceLanguage.PYTHON,
-                                sources))
-            else:
-                raise ValueError(
-                    "PROGRAMS not specified in catkin_install_python"
-                )
-    return src_info
+
+@attr.s(auto_attribs=True, slots=True)
+class NodeletExecutableInfo(ExecutableInfo):
```
ChrisTimperley, line 72: Is there need for both subclassing and the kind enum?

src/roswire/common/source.py
```diff
@@ -29,100 +32,70 @@ class SourceLanguage(enum.Enum):
...
+                                     SourceLanguage(info["language"]),
+                                     ExecutableKind(info["kind"]),
+                                     set(info["sources"]),
+                                     set(info["path_restrictions"]),
+                                     info['entrypoint'])
+
+
+@attr.s(auto_attribs=True, slots=True, frozen=True)
+class CMakeInfo:
+    executables: t.Mapping[str, ExecutableInfo]
```
ChrisTimperley, line 92: We should either: (a) rename this to `targets`, and/or (b) maintain two separate mappings (one for binaries and another for libraries)

src/roswire/common/source.py
```diff
@@ -134,6 +107,170 @@ def for_app_instance(
...
+        self,
+        args: t.List[str],
+        cmake_env: t.Dict[str, str],
+        executables: t.Dict[str, ExecutableInfo]
+    ) -> None:
+        opts, args = cmake_argparse(
+            args,
+            {"PROGRAMS": "*", "DESTINATION": "*"}
+        )
+        if 'PROGRAMS' in opts:
```
ChrisTimperley, line 257: prefer `not in` and then the code below can be unindented for better readability

src/roswire/common/source.py
```diff
@@ -134,6 +107,170 @@ def for_app_instance(
...
+            for i in range(len(opts['PROGRAMS'])):
```
ChrisTimperley, line 258: Why not `for program in opts["PROGRAMS"]`?

src/roswire/common/source.py
```diff
@@ -134,6 +107,170 @@ def for_app_instance(
...
+                # http://docs.ros.org/en/jade/api/catkin/html/howto/format2/installing_python.html  # noqa: F401, E501
+                # Convention is that ros python nodes are in nodes/ directory.
+                # All others are in scripts/. So just include python installs
+                # that are in nodes/
+                program = opts['PROGRAMS'][i]
+                if program.startswith("nodes/"):
+                    name = Path(program[0]).stem
+                    sources = set(program)
+                    if 'cwd' in cmake_env:
+                        sources = set(os.path.join(cmake_env['cwd'], program))
```
ChrisTimperley, line 268: If I remember correctly, this will produce a set of single-character strings obtained from iterating over the result of `os.path.join`.

src/roswire/ros2/source.py
```diff
@@ -31,22 +31,30 @@ def for_app_instance(
...
                 "Do not know how to process ROS2 packages with setup.py yet."
             )
-            return {}
+            raise NotImplementedError("Do not know how to process ROS2 "
+                                      "packages with setup.py yet.")
 
         logger.error(f"There is no package information inside "
                      f"{path_to_package}. Is it a package soure directory?")
-        return {}
+        raise ValueError(f"No pacakge information for {path_to_package}.")
```
ChrisTimperley, line 53: typo

src/roswire/common/source.py
schmerl: Done

src/roswire/common/source.py
schmerl: Good point. In a previous version of the code I needed `i`

src/roswire/common/source.py
schmerl: Fixed

src/roswire/ros2/source.py
schmerl: Done

src/roswire/common/nodelet_xml.py
```diff
@@ -0,0 +1,62 @@
+# -*- coding: utf-8 -*-
+import typing as t
+import xml.dom.minidom as dom
+
+import attr
+from loguru import logger
+
+
+@attr.s(frozen=True, auto_attribs=True)
+class NodeletLibrary:
```
ChrisTimperley, line 10: if it's not too much trouble, could you use a docstring to briefly describe these attributes?

src/roswire/common/source.py
```diff
@@ -29,111 +33,268 @@ class SourceLanguage(enum.Enum):
...
+@attr.s(auto_attribs=True, slots=True)
+class CMakeLibraryTarget(CMakeBinaryTarget):
+
+    _entrypoint: t.Optional[str] = attr.ib(init=False)
 
+    @property
+    def entrypoint(self) -> t.Optional[str]:
+        return self._entrypoint
+
+    @entrypoint.setter
```
ChrisTimperley, line 77: why would instances of this class be mutable?

src/roswire/common/source.py
```diff
@@ -29,111 +33,268 @@ class SourceLanguage(enum.Enum):
...
-    ) -> t.Mapping[str, NodeSourceInfo]:
+        package: Package
+    ) -> CMakeInfo:
         ...
+
+    @abc.abstractmethod
+    def package_paths(self, package: Package) -> t.Set[str]:
+        ...
+
+    def get_entrypoints(self, package: Package) -> t.Mapping[str, str]:
```
ChrisTimperley, line 127: suggest renaming to `get_nodelet_entrypoints` to be more specific

src/roswire/common/source.py
```diff
@@ -29,111 +33,268 @@ class SourceLanguage(enum.Enum):
...
+
+        Parameters
+        ----------
+        file_contents: str
+            The contents of the CMakeLists.txt file
+        package: Package
+            The package where the CMakeLists.txt file is defined
+        cmake_env: t.Dict[str, str]
+            Any context variables for processing the contents
+        source_extractor
```
ChrisTimperley, line 162: ?

src/roswire/common/source.py
```diff
@@ -29,111 +33,268 @@ class SourceLanguage(enum.Enum):
...
+
+        Returns
+        -------
+        t.Mapping[str, CMakeTarget]
```
ChrisTimperley, line 166: outdated

src/roswire/ros2/source.py
```diff
@@ -29,24 +26,32 @@ def for_app_instance(
...
+
+    def package_paths(self, package: Package) -> t.Set[str]:
+        # TODO Do this properly
+        include: str = os.path.normpath(
```
ChrisTimperley, line 54: what is this used for?

src/roswire/ros1/source.py
```diff
@@ -29,15 +24,22 @@ def for_app_instance(
...
 
         contents = self._files.read(cmakelists_path)
-        source_infos = extract_sources_from_cmake(contents)
-        return {n.node_name: n for n in source_infos}
+        return self._process_cmake_contents(contents, package, {})
+
+    def package_paths(self, package: Package) -> t.Set[str]:
+        # TODO Do this properly
+        include: str = os.path.normpath(
+            os.path.join(package.path, f'../../include/{package.name}')
```
ChrisTimperley, line 43: same as below: what's this used for?

src/roswire/ros2/source.py
schmerl: I think this takes paths with ../../. and gets rid of them.

src/roswire/ros1/source.py
schmerl: Sane as above

src/roswire/common/source.py
schmerl: Done.

src/roswire/common/nodelet_xml.py
schmerl: Done.

PR #471, opened 2021-09-08
Refactor API to ease ROSDiscover integration

schmerl:
This changes the API so that an app_instance is not required. (There is a connection to ROS already established in ROS2 but not in ROS1 of `App`.) The cmake parsing only requires `dockerblade.FileSystem` so we use that to build the CMakeParser.


ChrisTimperley (approved):
Looks good to me 👍🏻

PR #474, opened 2021-09-13
Added workspace include directories to package_paths in ROS1 source extractor

schmerl:


src/roswire/ros1/source.py
```diff
@@ -57,9 +57,41 @@ def get_cmake_info(
...
+            if self._files.exists(catkin_tools_dir):
+                result = workspace_path
+
+            workspace_path = os.path.dirname(workspace_path)
+
+        if result is None:
+            raise ValueError(f"unable to determine workspace for package: "
+                             f"{package}")
+
+        return {os.path.join(result, 'devel/include'),
```
ChrisTimperley, line 90: `os.path.join(result, 'devel/include', package.name)` etc.

ChrisTimperley: we probably want to check that these paths exist, too.

src/roswire/ros1/source.py
```diff
@@ -57,9 +57,41 @@ def get_cmake_info(
                 target.entrypoint = entrypoint
         return info
 
+    def _find_package_devel_workspaces(self, package: Package) -> t.Set[str]:
```
ChrisTimperley, line 60: The name and docs of the method don't quite make sense any more. Suggestion: keep the original `_find_package_workspace` function from ROSDiscover and move the finding of include directories to `package_paths`.

src/roswire/ros1/source.py
```diff
@@ -57,9 +57,44 @@ def get_cmake_info(
                 target.entrypoint = entrypoint
         return info
 
+    def _find_package_workspace(self, package: Package) -> str:
+        """Determines the absolute path of the workspace to which a given package belongs.
```
ChrisTimperley, line 61: nitpick: missing blank line to indicate a new docs section (Raises)

src/roswire/ros1/source.py
```diff
@@ -57,9 +57,44 @@ def get_cmake_info(
...
+        paths = {package.path}
+        workspace = self._find_package_workspace(package)
+
+        for contender in ('devel/include',
+                          'devel_isolated/include',
+                          'install/include',
+                          ):
+            workspace_contender = os.path.join(workspace, contender)
+            if self._files.exists(workspace_contender):
+                paths.add(workspace_contender)
```
ChrisTimperley, line 95: no need for these lines

src/roswire/ros1/source.py
```diff
@@ -57,9 +57,44 @@ def get_cmake_info(
...
+            workspace_contender = os.path.join(workspace_contender, package.name)
```
ChrisTimperley, line 96: we can flatten this into a single `os.path.join(workspace, contender, package.name)`

ChrisTimperley (approved):
These changes look good 👍🏻 Thanks!

PR #475, opened 2021-09-15
Make cmake commands case sensitive

schmerl:
The parser was not recognizing ADD_EXECUTABLE and add_executable as the same thing. This is fixed, and addresses: https://github.com/rosqual/rosdiscover/issues/184


ChrisTimperley (approved):
LGTM 👍🏻

PR #479, opened 2021-09-29
Process the source files in CMakeLists.txt to real files on the container

schmerl:
1. Process project to set `PROJECT_NAME`
2. Process set_target_directory to resolve node names that use `${PROJECT_NAME}`
3. Find files on the container that match the source prefix

### DIFF
--- a/src/roswire/common/source.py
+++ b/src/roswire/common/source.py
@@ -184,6 +184,9 @@ class CMakeExtractor(abc.ABC):
         context = ParserContext().parse(file_contents, skip_callable=False)
         for cmd, args, _arg_tokens, (_fname, _line, _column) in context:
             cmd = cmd.lower()
+            if cmd == "project":
+                opts, args = cmake_argparse(args, {})
+                cmake_env["PROJECT_NAME"] = args[0]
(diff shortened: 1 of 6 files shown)


src/roswire/common/source.py
```diff
@@ -303,13 +333,15 @@ def __process_python_executables(
...
-                    sources.add(program)
+                source = self._resolve_to_real_file(program, package, cmake_env)
+                sources.add(source)
                 logger.debug(f"Adding Python sources for {name}")
                 executables[name] = CMakeTarget(name,
                                                 SourceLanguage.PYTHON,
                                                 sources,
                                                 set())
+
+    def _list_to_dict(self, values: t.List[str]) -> t.Dict[str, str]:
```
ChrisTimperley, line 344: How does a list get converted to a dict? Seems like there's more going on here? It looks like the list is of the form: `[k1, v1, k2, v2, ...]`? Maybe rename this to `_key_val_list_to_dict`?

ChrisTimperley (approved):
Other than a minor comment about a method name, this looks great 👍🏻

PR #507, opened 2021-10-29
Convert node to list so that they are processed in order.

schmerl:
Nodes in a LaunchConfig were sets, which breaks the ordering required for starting nodelets after nodelet managers.

### DIFF
--- a/src/roswire/common/launch/config/launch.py
+++ b/src/roswire/common/launch/config/launch.py
@@ -11,6 +11,7 @@ from typing import (
     Mapping,
     Optional,
     Sequence,
+    Tuple,
 )
 
 import attr


src/roswire/common/launch/config/launch.py
```diff
@@ -31,7 +31,7 @@
 
 @attr.s(frozen=True, slots=True)
 class LaunchConfig:
-    nodes: Sequence[NodeConfig] = attr.ib(default=())
+    nodes: Tuple[NodeConfig, ...] = attr.ib(default=())
```
ChrisTimperley, line 34: why this change? (tuples are sequences)

schmerl:
```
src/roswire/common/launch/config/launch.py:140: error: Unsupported left operand type for + ("Sequence[NodeConfig]")
```

PR #509, opened 2021-11-01
Try to find the cmake target that matches a nodelet by class

schmerl:


ChrisTimperley (approved):
This looks good, but we should be careful to create an issue to keep track of this.

PR #515, opened 2021-11-06
Fixed error in reading contents of rosparam.

schmerl:
The following tag was giving a value of "" for rosparam.

```
launch>
  <node pkg='autorally_core' type='error_reproducer' name="error_reproducer">
    <rosparam param='topic'>
- direction: sub
  name: /left_camera
  type: sensor_msgs/Image
- direction: sub
  name: /right_camera
  type: sensor_msgs/Image
    </rosparam>
  </node>
</launch>
```

This also allows for unknown node executables. i.e., if we try to find a node that isn't on the container. In this case instead of raising an exception we report and error and continue.

### DIFF
--- a/src/roswire/ros1/launch/reader.py
+++ b/src/roswire/ros1/launch/reader.py
@@ -51,7 +51,7 @@ Loader = Callable[
 def _read_contents(tag: ET.Element) -> str:
     """Reads the text contents of an XML element."""
     # FIXME add support for CDATA -- possibly via lxml or xml.dom?
-    return tag.text | ""  # return "".join(t.text for t in tag if t.text)
+    return tag.text if tag.text else ""  # return "".join(t.text for t in tag if t.text)
 
 
 def _parse_bool(attr: str, val: str) -> bool:
--- a/src/roswire/ros1/launch/reader.py
+++ b/src/roswire/ros1/launch/reader.py
@@ -38,7 +38,7 @@ from ...name import (
 )
 
 if typing.TYPE_CHECKING:
-    from src import AppInstance
+    from ... import AppInstance
 
 _TAG_TO_LOADER = {}
 
--- a/src/roswire/ros1/launch/reader.py
+++ b/src/roswire/ros1/launch/reader.py
@@ -51,7 +51,7 @@ Loader = Callable[
 def _read_contents(tag: ET.Element) -> str:
     """Reads the text contents of an XML element."""
     # FIXME add support for CDATA -- possibly via lxml or xml.dom?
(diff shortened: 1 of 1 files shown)


src/roswire/ros1/launch/reader.py
```diff
@@ -372,7 +372,11 @@ def _load_node_tag(
 
         # locate node executable and determine the type
         executable_path = self.locate_node_binary(package, node_type)
-        executable_type = self._get_executable_type(executable_path)
+        if executable_path != "\\unknown":
```
ChrisTimperley, line 375: this seems unrelated?

src/roswire/ros1/launch/reader.py
```diff
@@ -652,7 +656,8 @@ def locate_node_binary(self, package: str, node_type: str) -> str:
                 f"unable to locate binary for node_type [{node_type}] "
                 f"in package [{package}]"
             )
-            raise ValueError(m)
+            logger.error(m)
+            return "\\unknown"
```
ChrisTimperley, line 660: is this going to break things elsewhere?

schmerl: Honestly, I don't know. But the error will be logged if it causes problems downstream (it only happens in the case where the node can't be found, and this used to just raise an error and crash).

src/roswire/ros1/launch/reader.py
schmerl: yeah, you're right.

schmerl: Changed comment to reflect this additional change.

src/roswire/ros1/launch/reader.py
```diff
@@ -51,7 +51,7 @@
 def _read_contents(tag: ET.Element) -> str:
     """Reads the text contents of an XML element."""
     # FIXME add support for CDATA -- possibly via lxml or xml.dom?
-    return "".join(t.text for t in tag if t.text)
+    return tag.text if tag.text else ""  # return "".join(t.text for t in tag if t.text)
```
ChrisTimperley, line 54: I worry a little bit about this breaking existing functionality, but not enough to block the changes before the paper deadline. Have you tried to see if we get the same results as before on existing systems? In any case, it makes sense to remove the commented-out code.

schmerl: I haven't checked...But it is definitely the case that `t.text for t in tag if t.text` doesn't work for the common use of rosparam.

ChrisTimperley (approved):
👍🏻 I think that we're good to merge

